PR #110, opened 2018-08-07
[feature] support custom plugins and providers

edulop91:
This pr does a couple of things:

1. Fogg did not generate a local plugin cache. Terraform assumes the cache directory exists or it will not cache plugins. The plugin directory is currently `<repo_root>/.terraform.d/plugin-cache/`

1. Similar to terraform plugin cache, we also create a custom terraform provider cache at `<repo_root>/terraform.d`. This contains custom terraform providers.

1. Similar to terraform plugin cache, we also create a bin plugin cache at `<repo_root>/.bin`. This contains executables we wish to make available to the docker image. Can be things like kubectl or other custom binaries. 

1. Fogg now has the ability to specify custom plugins from remote sources. Fogg will fetch the remote target and place it into a local directory (currently `<repo_root>/.bin/` and `<repo_root>/terraform.d/`. The custom plugin can then be vendored.
```json
{
  "plugins": {
    "terraform_providers": {
      "terraform-provider-bless": {
        "url": "https://github.com/chanzuckerberg/terraform-provider-bless/releases/download/v0.2.3/terraform-provider-bless_0.2.3_linux_amd64.tar.gz",
        "format": "tar"
      }
    },
    "custom_plugins": {
      "kubectl": {
        "url": "some url",
        "format": "tar"
      }  
    }
  }
}
```

Note: There is no caching mechanism for these yet and `fogg apply` can be somewhat slow.


apply/apply.go
```diff
@@ -96,8 +106,7 @@ func applyAccounts(fs afero.Fs, p *plan.Plan, accountBox *packr.Box) (e error) {
 	return nil
 }
 
-func applyModules(fs afero.Fs, p map[string]plan.Module, moduleBox *packr.Box) error {
```
edulop91: for consistency

apply/apply.go
```diff
@@ -62,14 +72,44 @@ func applyRepo(fs afero.Fs, p *plan.Plan, repoTemplates *packr.Box) error {
...
+		log.Infof("Applying custom provider %s", providerName)
+		err = customProvider.Install(providerName, fs)
+		if err != nil {
+			return errors.Wrapf(err, "Error applying custom provider %s", providerName)
+		}
+	}
+	return
+}
+
+// Terraform won't cache providers unless the directory structure already exists
```
ryanking: This is a good catch. Didn't realize this wasn't already happening.

apply/apply.go
```diff
@@ -62,14 +72,44 @@ func applyRepo(fs afero.Fs, p *plan.Plan, repoTemplates *packr.Box) error {
...
+func applyCustomPluginCache(fs afero.Fs, p *plan.Plan, repoBox *packr.Box) error {
+	path := path.Join("", providers.CustomPluginCacheDir)
```
ryanking: Why do we need to join here?

apply/apply.go
```diff
@@ -62,14 +72,44 @@ func applyRepo(fs afero.Fs, p *plan.Plan, repoTemplates *packr.Box) error {
...
 		}
 	}
-	return nil
+	return applyCustomProviders(fs, p)
+}
+
+func applyCustomProviders(fs afero.Fs, p *plan.Plan) (err error) {
+	for providerName, customProvider := range p.CustomProviders {
+		log.Infof("Applying custom provider %s", providerName)
+		err = customProvider.Install(providerName, fs)
```
ryanking: We usually put fs first, not sure if we should continue that style.

apply/apply.go
```diff
@@ -62,14 +72,44 @@ func applyRepo(fs afero.Fs, p *plan.Plan, repoTemplates *packr.Box) error {
...
+	err := fs.MkdirAll(path, 0755)
+	if err != nil {
+		return errors.Wrapf(err, "unable to make directory %s", path)
+	}
+	return applyTree(repoBox, fs, path, p.SiccMode, p)
+}
+
+// Terraform won't cache providers unless the directory structure already exists
+func applyPluginCache(fs afero.Fs, p *plan.Plan, repoBox *packr.Box) error {
+	path := path.Join("", providers.PluginCacheDir)
```
ryanking: Same where. Why join?

config/config.go
```diff
@@ -8,25 +8,27 @@ import (
...
-	AWSRegionBackend   string            `json:"aws_region_backend" validate:"required"`
-	AWSRegionProvider  string            `json:"aws_region_provider" validate:"required"`
-	AWSRegions         []string          `json:"aws_regions,omitempty"`
-	ExtraVars          map[string]string `json:"extra_vars"`
-	InfraBucket        string            `json:"infra_s3_bucket" validate:"required"`
-	Owner              string            `json:"owner" validate:"required"`
-	Project            string            `json:"project" validate:"required"`
-	TerraformVersion   string            `json:"terraform_version" validate:"required"`
+	AccountID          *int64                               `json:"account_id,omitempty"`
+	AWSProfileBackend  string                               `json:"aws_profile_backend,omitempty" validate:"required"`
```
ryanking: required + omitempty seems unnecessary, yeah?

providers/constants.go
```diff
@@ -0,0 +1,12 @@
+package providers
+
+const (
+	// CustomPluginCacheDir is the directory used by terraform to search for custom providers
+	// We default to linux_amd64 since we're running terraform inside of docker
+	// We vendor providers here
+	// See https://www.terraform.io/docs/configuration/providers.html#third-party-plugins
+	CustomPluginCacheDir = "terraform.d/plugins/linux_amd64"
+	// PluginCacheDir is the directory where terraform cahes tf approved providers
```
ryanking: s/cahes/caches/

providers/custom_provider.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return "", errors.Wrap(err, "could not create temporary directory")
+	}
+	resp, err := http.Get(cp.URL)
+	if err != nil {
+		return "", errors.Wrapf(err, "could not get %s", cp.URL)
+	}
+	defer resp.Body.Close()
+	_, err = io.Copy(tmpFile, resp.Body)
+	return tmpFile.Name(), errors.Wrap(err, "could not download file")
```
ryanking: I presume errors.Wrap will do the right thing if err is nil?

providers/custom_provider.go
```diff
@@ -0,0 +1,126 @@
...
+	case TypeProviderFormatTar:
+		return cp.processTar(path, fs)
+	default:
+		return errors.Errorf("Unknown provider format %s", cp.Format)
+	}
+}
+
+// https://medium.com/@skdomino/taring-untaring-files-in-go-6b07cf56bc07
+func (cp *CustomProvider) processTar(path string, dest afero.Fs) error {
+	err := util.CreateDirIfNotExists(CustomPluginCacheDir, dest)
```
ryanking: similar to above, the fs should probably be the first argument to CreateDirIfNotExists

util/dir.go
```diff
@@ -0,0 +1,16 @@
+package util
+
+import (
+	"github.com/pkg/errors"
+	"github.com/spf13/afero"
+)
+
+// CreateDirIfNotExists will create a directory (and all intermediate directories) if they do not exist
+func CreateDirIfNotExists(target string, dest afero.Fs) error {
```
ryanking: fs first I think

util/dir.go
```diff
@@ -0,0 +1,16 @@
...
+	if _, err := dest.Stat(target); err != nil {
+		if err := dest.MkdirAll(target, 0755); err != nil {
```
ryanking: Can you just call MkdirAll? I assume it does this stat thing already?

apply/apply.go
edulop91: was joining to `rootPath` and forgot to rm it when i got rid of it

providers/custom_provider.go
edulop91:
```
// Wrap returns an error annotating err with a stack trace
// at the point Wrap is called, and the supplied message.
// If err is nil, Wrap returns nil.
func Wrap(err error, message string) error {
	if err == nil {
		return nil
	}
```

apply/apply.go
```diff
@@ -88,23 +126,22 @@ func applyAccounts(fs afero.Fs, p *plan.Plan, accountBox *packr.Box) (e error) {
...
-	var e error
+func applyModules(fs afero.Fs, p map[string]plan.Module, moduleBox *packr.Box) (e error) {
 	for module, modulePlan := range p {
 		path := fmt.Sprintf("%s/modules/%s", rootPath, module)
 		e = fs.MkdirAll(path, 0755)
 		if e != nil {
 			return errors.Wrapf(e, "unable to make path %s", path)
 		}
-		e = applyTree(moduleBox, fs, path, modulePlan.SiccMode, modulePlan)
+		e = applyTree(fs, moduleBox, path, modulePlan.SiccMode, modulePlan)
```
ryanking: sorry I didn't realize this wasn't already consistent

apply/apply.go
```diff
@@ -52,24 +62,52 @@ func Apply(fs afero.Fs, conf *config.Config, tmp *templates.T, siccMode bool) er
...
+	return
+}
+
+func applyCustomPluginCache(fs afero.Fs, p *plan.Plan, repoBox *packr.Box) error {
+	path := providers.CustomPluginCacheDir
+	err := fs.MkdirAll(path, 0755)
+	if err != nil {
+		return errors.Wrapf(err, "unable to make directory %s", path)
+	}
+	return applyTree(fs, repoBox, path, p.SiccMode, p)
```
ryanking: Why does the repo template set need to be applied here?

apply/apply.go
```diff
@@ -52,24 +62,52 @@ func Apply(fs afero.Fs, conf *config.Config, tmp *templates.T, siccMode bool) er
...
+}
+
+// Terraform won't cache providers unless the directory structure already exists
+func applyPluginCache(fs afero.Fs, p *plan.Plan, repoBox *packr.Box) error {
+	path := providers.PluginCacheDir
+	err := fs.MkdirAll(path, 0755)
+	if err != nil {
+		return errors.Wrapf(err, "unable to make directory %s", path)
+	}
+	return applyTree(fs, repoBox, path, p.SiccMode, p)
```
ryanking: and here?

ryanking: note that if we just put a file in those directories in templates/repo it will do it for us. maybe .gitignore?

apply/apply.go
edulop91: hmm poorly named, renaming. 
```
	e = applyPluginCache(fs, p, &tmp.PluginCache)
	if e != nil {
		return errors.Wrap(e, "unable to apply plugin cache")
	}
	e = applyCustomPluginCache(fs, p, &tmp.CustomPluginCache)
	if e != nil {
		return errors.Wrap(e, "unable to apply custom plugin cache")
	}
```
they get called with their respective templates

PR #114, opened 2018-08-10
Refactor plan structs

edulop91:


plan/plan.go
```diff
@@ -8,64 +8,52 @@ import (
...
-	TerraformVersion   string
-	Type               string
+	AWSConfiguration
+	Components       map[string]Component
+	Env              string
+	ExtraVars        map[string]string
+	Owner            string
+	Project          string
+	TerraformVersion string
+	Type             string
```
edulop91: @ryanking should we now rm `Type` ?

ryanking: yes

PR #116, opened 2018-08-14
[breaking] Version docker image

edulop91:
Added a new field to the config `docker_image_version` to start versioning the chanzuckerberg/terraform docker image. 
Works with https://github.com/chanzuckerberg/images/pull/25


ryanking:
Do you think this should be something configured in fogg.json vs hard-coded in fogg? It seems like there will be a tight coupling between the makefiles generated by fogg and the image, yeah?

PR #123, opened 2018-08-21
Fix bless_ssh_config and add .update-readme.sh scripts

edulop91:


templates/module/.update-readme.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+# Auto-generated by fogg. Do not edit
+# Make improvements in fogg, so that everyone can benefit.
+
+# I would have written this diretly in the Makefile, but that was difficult.
```
mbarrien: s/diretly/directly/

PR #124, opened 2018-08-21
Plain files were being copied to top level, adjust the path

edulop91:


edulop91:
@mbarrien missed this in my prev pr

PR #130, opened 2018-09-04
de-dupe update-readme.sh

### LINKED ISSUE #66: move update-readme.sh to top level

ryanking:
To reduce noise, this moves to having a single script that we reference from each module Makefile.

Also adds abiltity to delete files that were previously managed by fogg.

Fixes #66


apply/apply.go
```diff
@@ -180,12 +180,21 @@ func applyTree(dest afero.Fs, source *packr.Box, targetBasePath string, subst in
 			if e != nil {
 				return errors.Wrapf(e, "unable to create file %s", target)
 			}
+		} else if extension == ".rm" {
+			_, e = dest.Stat(target)
```
edulop91: could also always os.remove and check the error type

ryanking: Oh yeah I forgot there is a "file doesn't exist" error type.

PR #132, opened 2018-09-05
move plan.Print to yaml.Marshal

ryanking:
Time to give up the custom thing and just let yaml.Marshal do its thing


plan/plan.go
```diff
@@ -138,111 +139,9 @@ func Eval(config *config.Config, verbose bool) (*Plan, error) {
...
-
-	fmt.Println("Modules:")
-	for name, module := range p.Modules {
-		fmt.Printf("\t%s:\n", name)
-		fmt.Printf("\t\tterraform_version: %s\n", module.TerraformVersion)
-	}
-	return nil
+	out, err := yaml.Marshal(p)
+	fmt.Print(string(out))
+	return err
```
edulop91: maybe wrap it

PR #133, opened 2018-09-05
Some docs around fogg.json config

edulop91:


docs/FAQ.md
```diff
@@ -6,4 +6,84 @@ When looking for a name we searched for people and things related to the concept
...
+
+## fogg.json
+
+### What are modules
+[Terraform modules](https://www.terraform.io/intro/getting-started/modules.html) are a collection of terraform resources (including module and data).
+
+### What are accounts
+Accounts correspond to AWS accounts. These can configure singleton resources such as IAM users and groups.
+
+### What are components
```
edulop91: this needs better wording

ryanking: there are a couple ways to explain it–

1. this is where you run apply
2. terraform calls these "root modules" we decided to give them a distinctive name to make it easier to talk about
3. they are like the main() for your tf code

Just some ideas.

PR #137, opened 2018-09-06
Remove TTY from all but fogg apply (closes #127)

### LINKED ISSUE #127: TTY should be optional
# Need
* As a developer using `fogg` in CI/CD pipelines, I want to use `make check-plan` to check the state of my terraform deployment, so that I know my infrastructure is properly deployed before deploying apps on top of it.

* As a developer using `fogg` in CI/CD pipelines, I want TTY to be optional in `fogg` Makefiles, so that I can run `fogg` make commands on non-TTY devices.

## Example
Running `fogg` make commands in a Gitlab pipeline will result in: `the input device is not a TTY`

mbarrien:
This PR removes all the `-it` flags from all docker calls except make apply, effectively removing the TTY allocation.

For the fogg apply, we create a new variable `FOGG_DOCKER_FLAGS` that defaults to blank except in the apply command. We separate apply into apply and real-apply, so that the `-it` flags only apply to the actual `terraform apply` command and not the fmt/get/init/ssh-forward targets, using make's support of [https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html](target-specific flags)

### DIFF
--- a/templates/repo/scripts/component.mk
+++ b/templates/repo/scripts/component.mk
@@ -46,10 +46,8 @@ get: ssh-forward
 plan: fmt get init ssh-forward
 	$(docker_terraform) plan
 
-apply: fmt get init ssh-forward real-apply
-
-real-apply: FOGG_DOCKER_FLAGS = -it
-real-apply:
+apply: FOGG_DOCKER_FLAGS = -it
+apply: fmt get init ssh-forward
 	$(docker_terraform) apply -auto-approve=false
 
 docs:
--- a/templates/repo/scripts/component.mk
+++ b/templates/repo/scripts/component.mk
@@ -10,12 +10,12 @@ TF=$(wildcard *.tf)
 IMAGE_VERSION=$(DOCKER_IMAGE_VERSION)_TF$(TERRAFORM_VERSION)
 
 docker_base = \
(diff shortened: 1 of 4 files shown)


templates/repo/scripts/component.mk
```diff
@@ -36,7 +36,10 @@ get: ssh-forward
 plan: fmt get init ssh-forward
 	$(docker_terraform) plan
 
-apply: fmt get init ssh-forward
+apply: fmt get init ssh-forward real-apply
```
edulop91: think this indirection might bring more complexity than benefit. 
running apply still requires the union of all its sub-command capabilities - you'd still need tty capabilities in order to run `apply` even if only allocated on `real-apply`.

templates/repo/scripts/component.mk
```diff
@@ -36,7 +36,10 @@ get: ssh-forward
...
+
+real-apply: FOGG_DOCKER_FLAGS = -it
```
edulop91: that's a neat make feature

templates/repo/scripts/component.mk
mbarrien: real-apply removed in latest. Please re-review and approve.

PR #144, opened 2018-09-12
[feature] check for version mis-match on apply

### LINKED ISSUE #51: warn when upgrading versions

ryanking:
This will check to see if the version in .fogg-version matches the current tool version. If --upgrade/-u is not supplied, apply will fail.

Currently the UX for this error is bad. After the PR I will do a pass at better error UX before we do a release.

Fixes #51

### DIFF
--- a/apply/apply.go
+++ b/apply/apply.go
@@ -30,9 +30,9 @@ const rootPath = "terraform"
 func Apply(fs afero.Fs, conf *config.Config, tmp *templates.T, upgrade bool) error {
 	if !upgrade {
 		toolVersion, _ := util.VersionString()
-		versionChange, _ := checkToolVersions(fs, toolVersion)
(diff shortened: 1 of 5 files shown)


apply/apply.go
```diff
@@ -56,6 +64,38 @@ func Apply(fs afero.Fs, conf *config.Config, tmp *templates.T) error {
 	return errors.Wrap(e, "unable to apply modules")
 }
 
+func checkToolVersions(fs afero.Fs, current string) (bool, string, error) {
+	f, e := fs.Open(".fogg-version")
+	if e != nil {
+		return false, "", errors.Wrap(e, "unable to open .fogg-version file")
+	}
+	reader := io.ReadCloser(f)
```
edulop91: not sure if you need this

PR #146, opened 2018-09-12
better error messages

### LINKED ISSUE #145: better error handling
For any error in fogg we currently issue a Panic and dump a stack trace. I think there are at least 3 buckets of errors that we should handle differently–

1. User errors - the user asked us to do something that we can't do. We should have a clear message on how to fix the problem and no stacktrace.
  * the config is invalid
  * they need to add `--upgrade` to `fogg apply`
2. Precondition errors - there is something about the current state of the repo which means we can't do our work. We should have a clear message that explains the problem and no stacktrace.
  * we can't read or write a file we need to
3. internal errors. We should have a message that something went wrong and dump information that would help use debug the issue (but nothing which couldn't be pasted into an issue here).
  * there is probably a bug in fogg and we would like the user to report it

I think we could have different Go types for each type of error and then some helper methods which create them. At the top level (cmd/root.go) we can have a switch which handles each differently.

Thoughts?

ryanking:
Fixes #145  and implements the scheme there with 1 change - rather than having 3 buckets, we group into 2–

1. Use errors – iow anything we'd expect a user to remediate

2. internal errors - things that are likely a bug

Also note that I didn't spend a ton of time thinking about whether each error should be User vs Internal and just defaulted to User for now for most of them.


Makefile
```diff
@@ -25,7 +28,7 @@ coverage: ## run the go coverage tool, reading file coverage.out
 	go tool cover -html=coverage.out
 
 test: packr ## run the tests
-	go test -cover ./...
+	gotest -cover ./...
```
ryanking: unrelated, but found it nice, esp when running in the vs code terminal

apply/apply.go
```diff
@@ -32,49 +33,54 @@ func Apply(fs afero.Fs, conf *config.Config, tmp *templates.T, upgrade bool) err
...
 
 	e = applyGlobal(fs, p.Global, &tmp.Global)
 	if e != nil {
-		return errors.Wrap(e, "unable to apply global")
+		return errs.WrapUser(e, "unable to apply global")
 	}
 
 	e = applyModules(fs, p.Modules, &tmp.Module)
-	return errors.Wrap(e, "unable to apply modules")
+	return errors.Wrap(e, "unable to apply modules") // FIXME
```
ryanking: This and a few other FIXMEs are here because when I change them things break (not the tests, but when running against an example repo). @edulop91 it is all related to plugins, maybe we can pair on fixing tomorrow?

apply/apply.go
```diff
@@ -97,40 +103,36 @@ func versionIsChanged(repo string, tool string) (bool, error) {
 }
 
 func applyRepo(fs afero.Fs, p *plan.Plan, repoTemplates *packr.Box) error {
-	e := applyTree(fs, repoTemplates, "", p)
-	if e != nil {
-		return e
-	}
-	return applyPlugins(fs, p)
+	return applyTree(fs, repoTemplates, "", p)
```
ryanking: liften this up to Apply

Makefile
```diff
@@ -6,6 +6,10 @@ LDFLAGS=-ldflags "-w -s -X github.com/chanzuckerberg/fogg/util.GitSha=${SHA} -X
 
 all: test install
 
+setup:
```
edulop91: :+1:

cmd/util.go
```diff
@@ -38,18 +37,20 @@ func readAndValidateConfig(fs afero.Fs, configFile string, verbose bool) (*confi
...
 	if err != nil {
 		fmt.Println("fogg.json has error(s):")
-		errs, ok := err.(validator.ValidationErrors)
+		validatonErrors, ok := err.(validator.ValidationErrors)
 		if ok {
-			for _, err := range errs {
-				msg := fmt.Sprintf("\t%s is a %s %s", err.Namespace(), err.Tag(), err.Kind())
-				fmt.Println(strings.Replace(msg, "Config.", "", 1))
+			var sb strings.Builder
+			for _, err := range validatonErrors {
```
edulop91: a project I sometimes use for this is https://github.com/hashicorp/go-multierror

util/template.go
```diff
@@ -24,12 +25,12 @@ func dict(in interface{}) map[string]interface{} {
...
+func OpenTemplate(source io.Reader) (*template.Template, error) {
 	s, err := ioutil.ReadAll(source)
 	if err != nil {
-		log.Panic(err)
+		return nil, errs.WrapInternal(err, "could not read template")
 	}
 	funcs := sprig.TxtFuncMap()
 	funcs["dict"] = dict
-	return template.Must(template.New("tmpl").Funcs(funcs).Parse(string(s)))
+	return template.Must(template.New("tmpl").Funcs(funcs).Parse(string(s))), nil
```
edulop91: maybe you can now rm the `template.Must` sinc eyou return an error

PR #147, opened 2018-09-13
gometalinter pass

ryanking:
A bunch of small fixes found by running `make lint-slow`.

### DIFF
--- a/.gometalinter.json
+++ b/.gometalinter.json
@@ -1,5 +1,7 @@
 {
   "Exclude": [
-    "exported (\\w+) ([\\w\\.]+) should have comment or be unexported"
+    "exported (\\w+) ([\\w\\.]+) should have comment or be unexported",
+    "templates/templates-packr.go",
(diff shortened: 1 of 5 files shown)


.gometalinter.json
```diff
@@ -1,5 +1,7 @@
 {
   "Exclude": [
-    "exported (\\w+) ([\\w\\.]+) should have comment or be unexported"
+    "exported (\\w+) ([\\w\\.]+) should have comment or be unexported",
+    "templates/templates-packr.go",
+    "error return value not checked \\(defer"
```
edulop91: i don't really know this dsl, but seems like `\\(defer` might be incomplete/typo

cmd/apply.go
```diff
@@ -66,9 +66,7 @@ var applyCmd = &cobra.Command{
 
 		// apply
 		e = apply.Apply(fs, config, templates.Templates, upgrade)
```
edulop91: return apply.Apply

PR #151, opened 2018-09-14
Add fogg apply/plan --no-plugins flag

mbarrien:
This PR scratches an itch about running fogg while offline or on a slow/expensive connection (e.g. from Caltrain)

It adds a flag to fogg plan and fogg apply, --no-plugins, which skips installing plugins. It's potentially a dangerous command, and it warns you there is danger, if you run terraform apply without having the most recent plugins installed, resulting in either commands failing, or in the worst case wrong behavior if a previously installed plugin had different behavior than the updated version. However, if you know for sure that your plugins are up to date, the flag provides an escape hatch to avoid the slow/expensive download that may not even be possible without Internet access.

It may not be the best way (a better way may be tracking a plugin cache), but wanted to at least start a discussion.

### DIFF
--- a/apply/apply.go
+++ b/apply/apply.go
@@ -28,7 +28,7 @@ import (
 const rootPath = "terraform"
 
 // Apply will run a plan and apply all the changes to the current repo.
-func Apply(fs afero.Fs, conf *config.Config, tmp *templates.T, upgrade bool) error {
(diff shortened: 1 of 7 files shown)


mbarrien:
@ryanking @edulop91 Please re-review

PR #167, opened 2018-10-10
[feature] Environment components can reference accounts remote state and get a map of account->accountID

edulop91:


templates/component/fogg.tf.tmpl
```diff
@@ -110,3 +110,16 @@ data "terraform_remote_state" "{{ $component }}" {
   }
 }
 {{ end }}
+
+{{ range $accountName, $account := .Accounts }}
+data "terraform_remote_state" "account_{{ $accountName }}" {
```
edulop91: added an account prefix, unsure either way

PR #176, opened 2018-10-18
Change release scripts - now run as `make release`

edulop91:


VERSION
```diff
@@ -1 +1 @@
-0.15.3
\ No newline at end of file
+0.16.0
```
edulop91: so this sort of thing stops happening

PR #181, opened 2018-10-29
Allow users to disable docker

edulop91:
Partial fix for #174 

Allows users to skip using docker by specifying `FOGG_DISABLE_DOCKER`. There is still work needed to:
* Make sure we install terraform and dependencies where needed
* Deal with custom plugins and providers

Should be no breaking changes in this pr


templates/repo/scripts/Makefile
```diff
@@ -0,0 +1,44 @@
...
+ifeq ($(shell uname -m),i686)
+  ARCH ?= "386"
+endif
+ifeq ($(shell uname -m),aarch64)
+  ARCH ?= "arm"
+endif
+
+TF_PRESENT := $(shell command -v terraform 2> /dev/null)
+TF_CURRENT_VERSION := $(shell terraform version | head -n1 | sed 's/Terraform v//g')
+TF_VERSION := 0.11.8
```
ryanking: this needs to come from a template var

templates/repo/scripts/Makefile
```diff
@@ -0,0 +1,44 @@
...
+
+TFLINT_PRESENT := $(shell command -v tflint 2> /dev/null)
```
ryanking: we can do this in a separate pr, but tflint should get installed like a plugin

templates/repo/scripts/Makefile
```diff
@@ -0,0 +1,44 @@
...
+endif
+
+tflint-required:
+ifndef TFLINT_PRESENT
+	$(error "tflint is not installed")
+endif
+
+install: install-terraform install-tflint
+
+install-terraform:
```
ryanking: should we consider https://github.com/Zordrak/tfenv ?

templates/repo/scripts/Makefile
edulop91: Oh yeah, pr still needs a bunch of work. Opened to get eyes on general approach

templates/repo/scripts/Makefile
edulop91: Yeah, still thinking how to reconcile plugins feature with no docker

ryanking: I haven't fully thought this through, but was thinking of managing a `bin` directory in the repo (or one for each arch) and adding that to the `PATH` in the makefiles.

edulop91: yeah, i was thinking something similar. will do in the next pr though so this doesn't get too big

templates/repo/scripts/component.mk
```diff
@@ -10,46 +10,57 @@ AUTO_APPROVE := false
...
 	-e RUN_USER_ID=$(shell id -u) -e RUN_GROUP_ID=$(shell id -g) \
 	-e TF_PLUGIN_CACHE_DIR="/repo/.terraform.d/plugin-cache" -e TF="$(TF)" \
 	-w /repo/$(REPO_RELATIVE_PATH) $(TF_VARS) $(FOGG_DOCKER_FLAGS) $$(sh $(REPO_ROOT)/scripts/docker-ssh-mount.sh)
-docker_terraform = $(docker_base) chanzuckerberg/terraform:$(IMAGE_VERSION)
-docker_sh = $(docker_base) --entrypoint='/bin/sh' chanzuckerberg/terraform:$(IMAGE_VERSION)
+docker_terraform := $(docker_base) chanzuckerberg/terraform:$(IMAGE_VERSION)
+docker_sh := $(docker_base) --entrypoint='/bin/sh' chanzuckerberg/terraform:$(IMAGE_VERSION)
+
+ifdef FOGG_DISABLE_DOCKER
+	sh_command ?= $(SHELL)
```
edulop91: or should we pick a shell?

ryanking: this is probably fine for now

PR #182, opened 2018-10-31
Some refactoring to move custom plugin logic to not depend on docker

edulop91:


edulop91:
getting rid of those in the next pr

PR #187, opened 2018-11-02
Experimental command to help generating aws configs

edulop91:


ryanking:
Rather than (or maybe in addition to) generating a config, perhas we could invoke `aws configure`? That would add it to the existing config file.

PR #188, opened 2018-11-02
snyk testing

ryanking:
This should fail the build if we introduce any insecure dependencies.


ryanking:
@edulop91 just the assume-role config

ryanking:
@edulop91 yeah

PR #191, opened 2018-11-16
[feature] generate .travis.yml

ryanking:
Auto-generate a .travis-ci.yml based on fogg.json.

This introduced a new top-level config in fogg.json for configuring things about travis ci.

When enabled it will use the accounts config to set up profiles and then bucket all the components and modules into a configured number of buckets.


plan/travisci.go
```diff
@@ -0,0 +1,72 @@
...
+type TravisCI struct {
+	Enabled          bool
+	AWSIDAccountName string
+	AWSProfiles      []AWSProfile
+	TestBuckets      [][]string
+}
+
+func (p *Plan) buildTravisCI(c *config.Config) TravisCI {
+	if p.Accounts == nil {
+		panic("buildTravisCI must be run after buildAccounts")
```
edulop91: why not return an error instead?

ryanking:
@edulop91 good idea

PR #193, opened 2018-11-21
Check for newer releases in Github

edulop91:


mbarrien:
Are we okay with having a silent remote connection every time we run fogg? Looks good otherwise.

PR #204, opened 2019-01-14
[feature] [breaking] fogg setup

ryanking:
In order to support running outside of docker containers, we have a new command `fogg setup` which will download and install custom plugins (and anything else we need in the future).

Note that this means "setup the current working copy of this repo" not "setup this repo".

This is a biggish change and includes a few yak-shaves–

1. new command `setup` led to some refactoring  of how we bootstrap the commands. There is probably room to DRY up the other commands, but will leave that to a follow-up.
2. The apply  command no longer does plugins so we no longer need the --no-apply-plugins flag.
3. We now template custom plugins stuff to inject the OS and Arch into them. Example of how to use this is [here](https://github.com/chanzuckerberg/shared-infra/compare/ryanking/symlink-spike?expand=1#diff-59e1b509adb3283f68ab8dce456995c2)

### DIFF
--- a/cmd/util.go
+++ b/cmd/util.go
@@ -14,9 +14,8 @@ import (
 	validator "gopkg.in/go-playground/validator.v9"
 )
 
-func openGitOrExit(pwd string) {
-	log.Debugf("opening git at %s", pwd)
-	_, err := os.Stat(".git")
+func openGitOrExit(fs afero.Fs) {
+	_, err := fs.Stat(".git")
 	if err != nil {
 		// assuming this means no repository
 		log.Fatal("fogg must be run from the root of a git repo")
@@ -67,4 +66,13 @@ func setupDebug(debug bool) {
 		logLevel = log.FatalLevel
 	}
 	log.SetLevel(logLevel)
+}
+
+func openFs() (afero.Fs, error) {
+	pwd, e := os.Getwd()
+	if e != nil {
+		return nil, e
+	}
+	fs := afero.NewBasePathFs(afero.NewOsFs(), pwd)
+	return fs, e
 }
--- a/cmd/util.go
+++ b/cmd/util.go
@@ -75,7 +75,7 @@ func openFs() (afero.Fs, error) {
 		return nil, e
 	}
 	fs := afero.NewBasePathFs(afero.NewOsFs(), pwd)
-	return fs, e
+	return fs, nil
 }
 
 func bootstrapCmd(cmd *cobra.Command, debug bool) (afero.Fs, *config.Config, error) {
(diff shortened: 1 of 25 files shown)


plugins/custom_plugin.go
```diff
@@ -99,29 +127,38 @@ func (cp *CustomPlugin) process(fs afero.Fs, pluginName string, path string, tar
 }
 
 func (cp *CustomPlugin) processBin(fs afero.Fs, name string, downloadPath string, targetDir string) error {
-	err := fs.MkdirAll(targetDir, 0755)
+	target, _ := Template(targetDir, runtime.GOOS, runtime.GOARCH)
```
edulop91: why ignore error?

plugins/custom_plugin.go
```diff
@@ -99,29 +127,38 @@ func (cp *CustomPlugin) process(fs afero.Fs, pluginName string, path string, tar
...
 		return errs.WrapUserf(err, "Could not move %s to %s", downloadPath, targetPath)
 	}
+
 	err = fs.Chmod(targetPath, os.FileMode(0755))
 	return errs.WrapUserf(err, "Error making %s executable", targetPath)
 }
 
 // https://medium.com/@skdomino/taring-untaring-files-in-go-6b07cf56bc07
 func (cp *CustomPlugin) processTar(fs afero.Fs, path string, targetDir string) error {
+	targetDir, _ = Template(targetDir, runtime.GOOS, runtime.GOARCH)
```
edulop91: why ignore error?

plugins/custom_plugin.go
ryanking: good catch

cmd/util.go
```diff
@@ -68,3 +68,40 @@ func setupDebug(debug bool) {
...
 	log.SetLevel(logLevel)
 }
+
+func openFs() (afero.Fs, error) {
+	pwd, e := os.Getwd()
+	if e != nil {
+		return nil, e
+	}
+	fs := afero.NewBasePathFs(afero.NewOsFs(), pwd)
+	return fs, e
```
edulop91: `fs, nil` I think

edulop91: (although e is nil)

plugins/custom_plugin.go
```diff
@@ -177,6 +220,8 @@ func (cp *CustomPlugin) processTar(fs afero.Fs, path string, targetDir string) e
 
 // based on https://golangcode.com/create-zip-files-in-go/
 func (cp *CustomPlugin) processZip(fs afero.Fs, downloadPath string, targetDir string) error {
+	targetDir, _ = Template(targetDir, runtime.GOOS, runtime.GOARCH)
```
edulop91: err being ignored

PR #205, opened 2019-01-24
[fix] ignore terraform.d to avoid checking in plugins

ryanking:


ryanking:
@mbarrien I pushed a change that should handle that.

PR #208, opened 2019-02-15
Add retries to .travis.yml install section as well

edulop91:


edulop91:
Have been seeing a fair amount of flakiness with these install steps

PR #210, opened 2019-04-09
[feature] Set us up to create components of types other than terraform; add a helm_template component kind

edulop91:


config/config.go
```diff
@@ -77,8 +102,10 @@ type Component struct {
 	AWSRegionBackend   *string           `json:"aws_region_backend"`
 	AWSRegionProvider  *string           `json:"aws_region_provider"`
 	AWSRegions         []string          `json:"aws_regions"`
+	EKS                *EKSConfig        `json:"eks,omitempty"`
```
mbarrien: It would be nice if the arguments to the components were pluggable based on which kind of component it is, but not gonna block a 1st pass at this.

mbarrien:
At a high level, I like this pluggable type system in fogg. I know there are specifics about the helm kind which I'm not sure what the right way to handle is (Helm "best practices" seem to prefer packaging Helm charts and releasing them as an artifact, not so much installing from a local directory. I ran into interesting behavior with respect to that along the way, specifically with respect to the versioning aspects of helm.)

edulop91:
@mbarrien I think this system still allows us to import packaged helm charts if you specify them in the requirements.yaml that gets generated

PR #211, opened 2019-04-09
Pass AWS environment variables to docker terraform

mat285:


ryanking:
@mat285 we are moving away from docker for running this. It is not well documented, but you can jus set `"docker": false` in your fogg.json.

mat285:
Awesome, thanks for the heads up! Is there any other docs I can take a look at?

ryanking:
@mat285 not really. I have a task do to a docs pass in our next sprint, so hopefully it will get better soon.

PR #218, opened 2019-04-16
Tar strip components when unpacking

edulop91:
helm tar comes packaged 1 level deep, use this to unbundle to the correct directory

### DIFF
--- a/plugins/custom_plugin.go
+++ b/plugins/custom_plugin.go
@@ -198,16 +198,15 @@ func (cp *CustomPlugin) processTar(fs afero.Fs, path string, targetDir string) e
 			return errs.NewUser("Nil tar file header")
 		}
 		// the target location where the dir/file should be created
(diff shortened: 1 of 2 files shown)


plugins/custom_plugin.go
```diff
@@ -192,7 +198,17 @@ func (cp *CustomPlugin) processTar(fs afero.Fs, path string, targetDir string) e
 			return errs.NewUser("Nil tar file header")
 		}
 		// the target location where the dir/file should be created
-		target := filepath.Join(targetDir, header.Name)
+		splitTarget := strings.Split(header.Name, string(os.PathSeparator))
+		fmt.Println(splitTarget)
```
ryanking: this should probably be log.Debug

edulop91: @ryanking forgot to rm them, were just here for dev checks

PR #220, opened 2019-04-16
Upgrade travisCI python version to fix TLS SNIMissingWarning errors

edulop91:


templates/travis-ci/.travis.yml.tmpl
```diff
@@ -1,13 +1,15 @@
...
 sudo: required
+python: 3.6
+
 {{ if .Docker }}
 services:
   - docker
 {{ end }}
 install:
-  - pip install awscli --upgrade --user
+  - pip install awscli --upgrade
```
edulop91: don't need the --user since running inside a virtualenv already

PR #221, opened 2019-04-17
Add missing targets to helm_template component makefile

edulop91:


templates/component/helm_template/Makefile.tmpl
```diff
@@ -25,3 +25,12 @@ set-eks-context:
 
 lint:
 	@helm lint -f variables.yaml .
+
+fmt:
```
edulop91: These are not well defined yet, so leaving a message and moving on. That way we don't break the top level `make fmt` etc

PR #230, opened 2019-05-02
config v2

ryanking:
This PR implements the structures and some validations for config format v2.


config/v2/config.go
```diff
@@ -0,0 +1,89 @@
...
+type Component struct {
+	common
+
+	EKS          *v1.EKSConfig     `json:"eks,omitempty"`
+	Kind         *v1.ComponentKind `json:"kind,omitempty"`
+	ModuleSource *string           `json:"module_source"`
+}
+
+type Providers struct {
+	AWS AWSProvider `json:"aws"`
```
edulop91: should that be optional `*AWSProvider`?

ryanking: Yeah, good point, need to find a way to mark that whole structure has not present (rather than just letting the fields get their zero-value).

config/v2/config.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+type AWSProvider struct {
+	AccountID         *int64   `json:"account_id"`
```
edulop91: what happens if this is not provided?

config/v2/config.go
```diff
@@ -0,0 +1,89 @@
...
+	AdditionalRegions []string `json:"additional_regions"`
+	Profile           *string  `json:"profile"`
+	Region            *string  `json:"region"`
+	Version           *string  `json:"version,omitempty"`
+}
+
+type Backend struct {
+	Bucket      string `json:"bucket,omitempty"`
+	DynamoTable string `json:"dynamodb_table,omitempty"`
```
edulop91: Is this required now? otherwise maybe make it a `*string`

PR #231, opened 2019-05-02
config conversion v1 -> v2

ryanking:
This code converts v1 structs to v2.


config/config.go
```diff
@@ -57,3 +58,161 @@ func FindAndReadConfig(fs afero.Fs, configFile string) (*v1.Config, error) {
...
 }
+
+type ver struct {
+	Version int `json:"version"`
+}
+
+func detectVersion(b []byte) (int, error) {
+	v := &ver{}
+	err := json.Unmarshal(b, v)
+	if err != nil {
```
edulop91: You can probably get away with not having this branch

config/config.go
```diff
@@ -57,3 +58,161 @@ func FindAndReadConfig(fs afero.Fs, configFile string) (*v1.Config, error) {
...
+		c2.Tools.TfLint = def1.TfLint
+	}
+
+	if c1.TravisCI != nil {
+		c2.Tools.TravisCI = c1.TravisCI
+	}
+
+	c2.Accounts = map[string]v2.Account{}
+
+	var ptrstr = func(s *string) string {
```
edulop91: was this defined already elsewhere in the code?

config/config.go
```diff
@@ -57,3 +58,161 @@ func FindAndReadConfig(fs afero.Fs, configFile string) (*v1.Config, error) {
...
+		if s == nil {
+			return ""
+		}
+		return *s
+	}
+
+	for acctName, acct := range c1.Accounts {
+		c2.Accounts[acctName] = v2.Account{
+			Common: v2.Common{
+				Backend: v2.Backend{
```
edulop91: If this will write the config back out to a file, might be worth ommiting these and instead let the inheritance chain take care of it

config/config.go
```diff
@@ -57,3 +58,161 @@ func FindAndReadConfig(fs afero.Fs, configFile string) (*v1.Config, error) {
...
+	Version int `json:"version"`
+}
+
+func detectVersion(b []byte) (int, error) {
+	v := &ver{}
+	err := json.Unmarshal(b, v)
+	if err != nil {
+		return 0, err
+	}
+	if v.Version == 0 {
```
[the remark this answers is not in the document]
ryanking: hashtag blame @nacin

config/config.go
ryanking: Yeah i think so, but i didn't bother refactoring because I'm not sure what we want to do about pointers.

config/config.go
ryanking: Because if err !=nil then version == 0?

config/config.go
ryanking: I'm not sure I follow - if someone set these values at the account level (presumably overriding the defaults) then we should preserve them, yeah?

config/config.go
edulop91: can return the `calculatedVersion,err` even if there's an error

config/config.go
edulop91: Actually, not sure what the behavior of `omitempty` is on a zero struct and this might actually be good. Was thinking if none of `acct.InfraBucket`,`InfraDynamoTable`, etc are defined then we might not want `v2.acct.Backend` to show up in the serialized config we write back out to a file

ryanking: Good points. I think I just need to watch out for this when I write the code to migrate config files. I'm sure some tweaking will be required.

config/config.go
ryanking: Ah yes, I was worried about `v` being nil, but that can't be the case.

PR #241, opened 2019-05-13
use afero for symlinks + switch to temporary directories for tests

ryanking:


apply/apply.go
```diff
@@ -425,16 +426,23 @@ func getTargetPath(basePath, path string) string {
...
+func linkFile(fs afero.Fs, name, target string) error {
 	log.Debugf("linking %s to %s", name, target)
+	linker, ok := fs.(afero.Symlinker)
+
+	if !ok {
+		return errs.NewInternal("can't cast to afero.SymLinker")
+	}
+
+	log.Debugf("removing link at %s", name)
+	fs.Remove(name)
```
edulop91: maybe log.debug this error

PR #242, opened 2019-05-13
golden file tests

ryanking:
As a way to tackle #87, #88 , etc, this introduces some "golden file tests" ([Background](https://medium.com/soon-london/testing-with-golden-files-in-go-7fccc71c43d3)).

The basic idea is that we have fogg.json files in testdata and can run apply over them and check them in (`make update-golden-files`). Then when we run the tests we apply to a temp directory and compare the results.

The code for comparing has some weaknesses, but is a starting point for making sure we don't introduce (unknown) changes to the outputted files.


testdata/v1_full/terraform/accounts/foo/terraform.d
```diff
@@ -0,0 +1 @@
+/Users/rking/go/src/github.com/chanzuckerberg/fogg/testdata/v1_full/terraform.d
```
edulop91: @ryanking is this going to be a problem?

util/module_storage.go
```diff
@@ -9,13 +9,24 @@ import (
...
-		return "", errs.WrapUser(e, "could not get pwd")
+func DownloadModule(fs afero.Fs, cacheDir, source string) (string, error) {
+
+	// We want to do these operations from the root of our working repository.
+	// In the case where we have a BaseFs we pull out its root. Otherwise use `pwd`.
+	var pwd string
+	if baseFs, ok := fs.(*afero.BasePathFs); ok {
+		pwd = afero.FullBaseFsPath(baseFs, ".")
+	} else {
+		var e error
```
edulop91: maybe bring this up to the top scope as well

util/module_storage_test.go
```diff
@@ -2,24 +2,37 @@ package util
...
 
-	d, e := DownloadModule(dir, "./test-module")
-	assert.Nil(t, e)
-	assert.NotNil(t, d)
-	assert.NotEmpty(t, d)
+	pwd, e := os.Getwd()
+	a.NoError(e)
+
+	fs := afero.NewBasePathFs(afero.NewOsFs(), pwd)
+	d, e := DownloadModule(fs, dir, "github.com/chanzuckerberg/fogg-test-module")
```
edulop91: does this mean you need network to be able to run tests now? do we want to add something like a `// +build integration` tag to files that require network?

ryanking:
As a way to tackle #87, #88 , etc, this introduces some "golden file tests" (Background).

The basic idea is that we have fogg.json files in testdata and can run apply over them and check them in (make update-golden-files). Then when we run the tests we apply to a temp directory and compare the results.

The code for comparing has some weaknesses, but is a starting point for making sure we don't introduce (unknown) changes to the outputted files.

PR #244, opened 2019-05-14
separate out tests that require internet

ryanking:


apply/golden_file_test.go
```diff
@@ -1,3 +1,5 @@
+// +build !online
```
edulop91: shouldn't it be `offline`?

ryanking: I just pushed a diff to invert this. Now it is `!offline` meaning "build this if we haven't set the offline tag".

PR #245, opened 2019-05-14
refactor plan to match config

ryanking:
### Summary
This is a refactoring to make it easier to make AWS  providers optional (and add other providers).

### Test Plan
All existing tests pass, plus I tested manually on shared-infra.


apply/golden_file_test.go
```diff
@@ -29,6 +30,7 @@ func TestIntegration(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.fileName, func(t *testing.T) {
 			a := assert.New(t)
+			r := require.New(t)
```
edulop91: huh, seems like this is what we'd want to be using in general rather than assert

PR #246, opened 2019-05-15
lint pass

ryanking:
Switch to golangci-lint + clean up findings.

### Summary

This started as a pass to clean up any lint findings, but found that I also needed to switch from gometalinter (which is slow and deprecated) to golangci-lint.

### Test Plan

`make lint test`

### References

* https://github.com/alecthomas/gometalinter
* https://github.com/golangci/golangci-lint


ryanking:
@edulop91 it would be great if we could just run https://golangci.com/ and have it comment on the PR directly

PR #247, opened 2019-05-15
[breaking] make aws provider optional

ryanking:
### Summary
This change makes it optional to specify an aws provider. Only planning on supporting this for config
v2 but it might accidentally work in v1.

The only breaking change is that we now require an aws account id for the global component, where before we would silently ignore the fact that it was missing. I consider this a bug / oversight so fixing it here.

### Test Plan

This diff adds a number of tests, especially golden file tests. None of the existing golden files had to be changed.

### References

*


apply/apply_test.go
```diff
@@ -218,9 +219,9 @@ func TestCreateFileNonExistentDirectory(t *testing.T) {
 }
 
 func TestApplySmokeTest(t *testing.T) {
-	a := assert.New(t)
+	r := require.New(t)
```
ryanking: require fails fast, while assert keeps going

apply/apply_test.go
```diff
@@ -231,6 +232,7 @@ func TestApplySmokeTest(t *testing.T) {
     "aws_profile_provider": "prof",
     "aws_profile_backend": "prof",
     "aws_provider_version": "0.12.0",
+    "account_id": 789,
```
ryanking: You now need to specify an account_id for global

config/v2/resolvers.go
```diff
@@ -0,0 +1,188 @@
+package v2
```
ryanking: this is mostly just moved code from v2/validations.go

config/v2/resolvers.go
```diff
@@ -0,0 +1,188 @@
...
+		for k, v := range m {
+			resolved[k] = v
+		}
+	}
+	return resolved
+}
+
+// ResolveAWSProvider will return an AWSProvder iff one of the required fields is set somewhere in the set of Common
+// config objects passed in. Otherwise it will return nil.
+func ResolveAWSProvider(commons ...Common) *AWSProvider {
```
ryanking: this is new tho

templates/component/terraform/fogg.tf.tmpl
```diff
@@ -10,14 +10,14 @@
   }
 
   # Aliased Providers (for doing things in every region).
-  {{ $out := . }}
+  {{ $outter := . }}
```
ryanking: trying to make it clearer that we are capturing the outer scope

testdata/v1_full/terraform/accounts/bar/terraform.d
```diff
@@ -1 +1 @@
-/Users/rking/go/src/github.com/chanzuckerberg/fogg/testdata/v1_full/terraform.d
\ No newline at end of file
+../../../terraform.d
```
ryanking: this was a bug, fixed now. unfortunately our tests don't catch this because we don't do any assertions on links

config/v2/config.go
```diff
@@ -135,12 +135,15 @@ func (c *Config) Generate(r *rand.Rand, size int) reflect.Value {
 	}
 
 	randAWSProvider := func(r *rand.Rand, s int) *AWSProvider {
-		return &AWSProvider{
-			AccountID: randInt64Ptr(r, size),
-			Region:    randStringPtr(r, s),
-			Profile:   randStringPtr(r, s),
-			Version:   randStringPtr(r, s),
+		if r.Float32() < 0.5 {
```
edulop91: I wonder if this method should be renamed from `Generate` to `GenerateTestConfig` to mark explicitly

config/v2/resolvers.go
```diff
@@ -0,0 +1,188 @@
...
+		t := getter(c)
+		if t != nil {
+			s = t
+		}
+	}
+	return s
+}
+
+// ResolveRequiredString will resolve the value and panic if it is nil. Only to be used after validations are run.
+func ResolveRequiredString(getter func(Common) *string, commons ...Common) string {
```
edulop91: I feel like for these the `Must...` is a kind of a go convention. Maybe `MustResolveString`

config/v2/resolvers.go
```diff
@@ -0,0 +1,188 @@
...
+			// optional fields
+			AccountID:         lastNonNilInt64(AWSProviderAccountIdGetter, commons...),
+			AdditionalRegions: ResolveOptionalStringSlice(AWSProviderAdditionalRegionsGetter, commons...),
+		}
+	}
+	return nil
+}
+
+func OwnerGetter(comm Common) *string {
+	return comm.Owner
```
edulop91: Maybe this can be comm.GetOwner()?

edulop91: Same with the others

config/v2/validation.go
```diff
@@ -41,229 +41,80 @@ func (c *Config) Validate() error {
...
-	}
-	return map[string]string{}
-}
+func (c *Config) ValidateAWSProviders() error {
+	var errs *multierror.Error
 
-func nonZeroInt64(i int64) bool {
-	return true
-}
+	var validate = func(p AWSProvider, component string) error {
```
edulop91: I wonder if should be `AWSProvider.Validate() error`

config/v2/validation.go
```diff
@@ -41,229 +41,80 @@ func (c *Config) Validate() error {
...
 	// For each account, we need the field to be valid in either the defaults or account
-	for acctName, acct := range c.Accounts {
-		v := lastNonNil(getter, c.Defaults.Common, acct.Common)
+	for name, acct := range c.Accounts {
+		v := ResolveAWSProvider(c.Defaults.Common, acct.Common)
 		if v == nil {
-			err = multierror.Append(err, fmt.Errorf("account %s must have field %s at either the account or defaults level", acctName, fieldName))
-		} else if !validator(*v) {
-			err = multierror.Append(err, fmt.Errorf("account %s must have a valid %s set at either the account or defaults level", acctName, fieldName))
+			// nothing to validate
```
edulop91: I feel like these branches should live inside the validate function

config/v2/resolvers.go
ryanking: Good idea. This diff is already getting big, so I would like to do this as a followup

config/v2/resolvers.go
ryanking: Yeah though we use `func (comm Common) *string` in a bunch of places.

config/v2/validation.go
ryanking: I am refactoring this a bit i the next PR (for snowflake provider). It is useful to have it as a function to keep the code generic.

Otherwise I think to take this approach I would have to create an `Validatable` interface or some such.

config/v2/validation.go
ryanking: working on it in the next PR

config/v2/config.go
ryanking: Yeah probably.

ryanking: Actually we can't do that. test/quick relies on this method name

ryanking: https://github.com/chanzuckerberg/fogg/issues/249

config/v2/resolvers.go
ryanking: https://github.com/chanzuckerberg/fogg/issues/250

PR #252, opened 2019-05-17
snowflake provider support

### LINKED ISSUE #225: fogg support for snowflake provider

ryanking:
### Summary
This adds support for configuring snowflake providers via fogg.

It also includes a few refactorings, most notably there is now a "common" set of templates which can be used to share template code across the various parts of fogg. The first thing added there is the snipped for configuring a snowflake provider, but over time we should be able to DRY up our templates quite a bit by moving things here.

### Test Plan

Added some golden file tests, manually inspected. Ran on a branch of our snowflake infra repo.

### References
* Fixes #225


config/v2/validation.go
```diff
@@ -42,68 +42,94 @@ func (c *Config) Validate() error {
...
+	return errs
+}
+
+func (c *Config) ValidateSnowflakeProviders() error {
+	var errs *multierror.Error
+
+	c.WalkComponents(func(component string, comms ...Common) {
+		v := ResolveSnowflakeProvider(comms...)
+		if v == nil {
+			// nothing to validate
```
edulop91: These conditionals make me think there is a better validation abstraction

config/v2/resolvers.go
```diff
@@ -99,6 +100,21 @@ func ResolveAWSProvider(commons ...Common) *AWSProvider {
 	return nil
 }
 
+func ResolveSnowflakeProvider(commons ...Common) *SnowflakeProvider {
+	account := lastNonNil(SnowflakeProviderAccountGetter, commons...)
+	role := lastNonNil(SnowflakeProviderRoleGetter, commons...)
+	region := lastNonNil(SnowflakeProviderRegionGetter, commons...)
+
+	if account != nil || role != nil || region != nil {
```
edulop91: below we make the assumption that if one is set then all are

plan/plan.go
```diff
@@ -263,6 +270,16 @@ func resolveComponentCommon(commons ...v2.Common) ComponentCommon {
 		}
 	}
 
+	var snowflakePlan *SnowflakeProvider
+	snowflakeConfig := v2.ResolveSnowflakeProvider(commons...)
+	if snowflakeConfig != nil {
+		snowflakePlan = &SnowflakeProvider{
+			Account: *snowflakeConfig.Account,
```
edulop91: This assumes if one is present then all are, but resolve doesnt

templates/common/snowflake_provider.tmpl
```diff
@@ -0,0 +1,9 @@
+{{define "snowflake_provider"}}
```
edulop91: 👍

config/v2/resolvers.go
ryanking: You mean below in this func? I don't think that's the case.

The current design thinking here is that the "resolution" logic will return the best version of the data we have, though it might be incomplete. Then validation will ensure that it is valid. Once we ensure that its valid we can confer the configs to plans knowing that all preconditions are met.

config/v2/validation.go
ryanking: That's almost certainly true, but I haven't come up with it yet.

plan/plan.go
ryanking: The trick here is that validation should assure that this is the case. And the rule is that you can only run the plan after validation. That said, there might be bugs...

PR #270, opened 2019-05-29
[feature] Add a command to migrate from v1 config to v2 configs

edulop91:
### Summary
`fogg upgrade`

### Test Plan
Say unittests

### References


cmd/v2upgrade.go
```diff
@@ -0,0 +1,90 @@
...
+	v1 "github.com/chanzuckerberg/fogg/config/v1"
+	"github.com/chanzuckerberg/fogg/errs"
+	"github.com/chanzuckerberg/fogg/plan"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	v2upgrade.Flags().StringP("config", "c", "fogg.json", "Use this to override the fogg config file.")
+	rootCmd.AddCommand(v2upgrade)
```
ryanking: I think we should just call this "upgrade" or something, so that it is more generic and we can add more stuff to it over time.

cmd/v2upgrade.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+
+		// check that we are at root of initialized git repo
+		openGitOrExit(fs)
+
+		bytes, version, err := config.FindConfig(fs, configFile)
+		if err != nil {
+			return err
+		}
+		switch version {
```
ryanking: I think this core logic probably should be moved to the config package. That should make it easier to test.

config/config.go
```diff
@@ -35,25 +35,33 @@ func InitConfig(project, region, bucket, table, awsProfile, owner, awsProviderVe
 	}
 }
 
-func FindAndReadConfig(fs afero.Fs, configFile string) (*v2.Config, error) {
+// FindConfig loads a config into memory
```
ryanking: ...and the config version

config/config.go
```diff
@@ -184,6 +189,16 @@ func UpgradeConfigVersion(c1 *v1.Config) (*v2.Config, error) {
 				TerraformVersion: env.TerraformVersion,
 			},
 		}
+
+		if env.InfraBucket != nil || env.InfraDynamoTable != nil || env.AWSProfileBackend != nil || env.AWSRegionBackend != nil {
```
edulop91: we do this so we don't end up including a bunch of `bucket: {}` in the output

config/upgrade_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/chanzuckerberg/fogg/util"
+)
+
+func TestUpgradeV2(t *testing.T) {
+	r := require.New(t)
+	confPath := "config"
+	conf := []byte(`{"version": 1}`)
+	fs, _, err := util.TestFs()
+	r.Nil(err)
```
ryanking: Not sure if it matters, but this might be better as NoError

PR #277, opened 2019-06-03
add ability to dump aws/config to stdout

ryanking:
For running in Atlantis, we need to be able to generate an entire aws/config file non-interactively.

This PR will allow you to do that with fogg exp aws-config -a -e -r rolename

It also removes the code for generating the profile session name, since that seems to be a dead-end in terms of enforcement.

### Test Plan
* unit tests, manual test against shared-infra repo


mbarrien:
Looks okay but needs a rebase against the Go modules PR I just merged.

PR #283, opened 2019-06-10
WIP: Yml migrate

shedninja6:
### Summary
- **apply**: added flag for fogg.yml and changed command line descriptions to match
- **aws_config**: Added flag for fogg.yml
- **config**: remove .json code and currently only supports yaml file **will need guidance on how to add json and yaml support**
- Created an integration test for YAML file. The test is a copy of the original .json integration test
- fogg.yml to testdata, in order to pass tests
- Made minor changes to comments, so that they show the support of .json and .yml file extensions

### Test Plan
Currently failing plan_tests because the tests rely on fogg.json file structure. Currently adding support for fogg.yml


apply/golden_file_test.go
```diff
@@ -119,100 +119,100 @@ func TestIntegration(t *testing.T) {
 	}
 }
 
-func TestIntegrationYaml(t *testing.T) {
```
ryanking: You shouldn't need to create a new go test, juts new test directories in `testdata`

cmd/exp/aws_config.go
```diff
@@ -18,11 +18,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
-//TODO: Potentially revert change
+//TODO: Enable aws_config.go to update yaml and json files
```
ryanking: this command should only have to read the files, right?

cmd/plan.go
```diff
@@ -10,15 +10,14 @@ import (
...
-	planCmd.Flags().StringP("config", "c", "fogg.json", "Use this to override the fogg config file.")
 	planCmd.Flags().StringP("config", "c", "fogg.yml", "Use this to override the fogg config file.")
 	rootCmd.AddCommand(planCmd)
 }
 
 var planCmd = &cobra.Command{
 	Use:           "plan",
 	Short:         "Run a plan",
-	Long:          "plan will read fogg.yml or fogg.json, use that to generate a plan and print that plan out. It will make no changes.",
+	Long:          "plan will read fogg.yml, use that to generate a plan and print that plan out. It will make no changes.",
```
ryanking: I think for the time being we should support both.

cmd/upgrade.go
```diff
@@ -10,7 +10,6 @@ import (
 )
 
 func init() {
-	upgrade.Flags().StringP("config", "c", "fogg.json", "Use this to override the fogg config file.")
```
ryanking: Even when we are yaml-only we should still allow people to specify a config file.

config/v1/config_test.go
```diff
@@ -83,7 +94,7 @@ func TestParse(t *testing.T) {
 	assert.NotNil(t, c.Modules)
 }
 
-func TestJsonFailure(t *testing.T) {
+func TestYamlFailure(t *testing.T) {
```
ryanking: rather than replace json, let's do both for now

shedninja6: If I want to test both I need to make a separate ReadConfig method inside config.go. Is that fine or is there a better way of going around that? @ryanking

PR #296, opened 2019-06-14
[feature] YAML support

shedninja6:
#### Main Additions
- **init**: Generates .yml file by default
- **apply**: Reads .yml files (Default is .yml but also support .json)
- **plan**: Reads and writes using .yml (Default is .yml but also support .json)
- **aws-config**: Supports .yml and .json
- **tests**: All tests are configured to test both .json and .yml (excluding golden_tests) 

#### Current PR Specifics
- Added yaml annotations to config structs
- Golden tests support .json and can support .yml but do not have yaml folders
- config.go tests will work for .json and .yml files
- config_test.go supports .json and .yml file reading
- created test yaml folder in testdata
- testing.go TestFile supports _yaml file paths

### Test Plan
- All unit tests current test .json and .yml
- No current unit tests for migration. I plan on adding tests to see if .yml and .json are supported after the migration


apply/golden_file_test.go
```diff
@@ -94,6 +113,7 @@ func TestIntegration(t *testing.T) {
 						r.NotNil(i1)
 						r.NoError(e1)
 
+						fmt.Println(path)
```
edulop91: rm

cmd/exp/yaml_migrate.go
```diff
@@ -0,0 +1,19 @@
+package exp
+
+import (
+	"github.com/chanzuckerberg/fogg/exp/yaml_migrate"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	ExpCmd.AddCommand(yamlMigrateCmd)
```
edulop91: `exp` subcommand was intended for experimental commands, this migration should be something we support more broadly. there's already an upgrade command, i wonder if we can latch onto that?

config/config.go
```diff
@@ -37,10 +38,21 @@ func InitConfig(project, region, bucket, table, awsProfile, owner, awsProviderVe
 
 // FindConfig loads a config and its version into memory
 func FindConfig(fs afero.Fs, configFile string) ([]byte, int, error) {
+	switch configFile {
+	case "fogg.json":
```
edulop91: does this depend on users not using the command line flag to change the default config?

config/config.go
```diff
@@ -57,25 +69,63 @@ func FindConfig(fs afero.Fs, configFile string) ([]byte, int, error) {
 	return b, v, nil
 }
 
+func findConfig(fs afero.Fs, configFile string) ([]byte, int, error) {
```
edulop91: i'm probably missing something here, but are `findJSON` and `findConfig` different?

config/config.go
```diff
@@ -57,25 +69,63 @@ func FindConfig(fs afero.Fs, configFile string) ([]byte, int, error) {
...
 		return nil, err
 	}
+
 	switch v {
-	case 1:
-		c, err := v1.ReadConfig(b)
-		if err != nil {
-			return nil, err
+	case 1: //Upgrade the config version
+		if strings.Compare(configFile, "fogg.json") == 0 {
```
edulop91: I think this also assumes that users have not overwritten the default fogg.json cli flag

config/config.go
```diff
@@ -57,25 +69,63 @@ func FindConfig(fs afero.Fs, configFile string) ([]byte, int, error) {
...
+			c, err := v1.ReadJsonConfig(b)
+			if err != nil {
```
edulop91:
```
			if err != nil {
				return nil, err
			}
			c2, err := UpgradeConfigVersion(c)
			return c2, err
```
is duplicated for both branches of this conditional, is there a way to avoid this duplication?

config/config.go
```diff
@@ -57,25 +69,63 @@ func FindConfig(fs afero.Fs, configFile string) ([]byte, int, error) {
...
 		}
-
-		c2, err := UpgradeConfigVersion(c)
-		return c2, err
-
 	case 2:
 
-		c2, err := v2.ReadConfig(b)
-		return c2, err
+		if strings.Compare(configFile, "fogg.json") == 0 {
```
edulop91: same as above here

config/v1/config.go
```diff
@@ -91,57 +92,68 @@ const (
...
+	Defaults Defaults           `json:"defaults" yaml:"defaults"`
+	Docker   bool               `json:"docker,omitempty" yaml:"docker,omitempty"`
+	Envs     map[string]Env     `json:"envs" yaml:"envs"`
+	Modules  map[string]Module  `json:"modules" yaml:"modules"`
+	Plugins  Plugins            `json:"plugins,omitempty" yaml:"plugins"`
+	TravisCI *TravisCI          `json:"travis_ci,omitempty" yaml:"travis_ci,omitempty"`
 }
 
 func ReadConfig(b []byte) (*Config, error) {
+	c := &Config{
```
edulop91: I wonder if readConfig should either know how to deal with both json and yaml files or print an error and ask users to migrate to yaml if it encounters a json file.

found this little blog post regarding how k8s deals with both json and yaml configuration, maybe there are some good ideas there https://sharpend.io/blog/teaching-go-programs-to-love-json-and-yaml/

config/v2/config.go
```diff
@@ -8,101 +8,116 @@ import (
...
+	"gopkg.in/yaml.v2"
 )
 
 func ReadConfig(b []byte) (*Config, error) {
 	c := &Config{
 		Docker: false,
 	}
-	e := json.Unmarshal(b, c)
+	e := yaml.Unmarshal(b, c)
+	fmt.Println(string(b))
```
edulop91: rm

config/v2/config.go
```diff
@@ -8,101 +8,116 @@ import (
...
+	if e != nil {
+		return nil, errs.WrapUser(e, "unable to parse yaml config file")
+	}
+	return c, nil
```
edulop91: you can just `return c, errs.WrapUser(e, ...)`. to avoid having that extra conditional. more conditionals == more complexity == more bugs == need to write more tests

config/v2/config.go
```diff
@@ -8,101 +8,116 @@ import (
...
+}
 
-	return c, errs.WrapUser(e, "unable to parse config")
+func ReadJsonConfig(b []byte) (*Config, error) {
+	c := &Config{
+		Docker: false,
+	}
+	e := json.Unmarshal(b, c)
+	if e != nil {
```
edulop91: same here

config/v2/config.go
```diff
@@ -8,101 +8,116 @@ import (
...
+		return nil, errs.WrapUser(e, "unable to parse yaml config file")
+	}
+	return c, nil
+}
 
-	return c, errs.WrapUser(e, "unable to parse config")
+func ReadJsonConfig(b []byte) (*Config, error) {
+	c := &Config{
+		Docker: false,
```
edulop91: seems like some are `docker: true` and others `docker: false` is there a reason for the inconsistency?

exp/yaml_migrate/yaml_migrate.go
```diff
@@ -0,0 +1,58 @@
...
+	"io/ioutil"
+	"os"
+
+	"github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v2"
+)
+
+//ConvertToYaml method converts fogg.json to fogg.yml
+func ConvertToYaml() error {
+	jsonFile, err := os.Open("fogg.json")
```
edulop91: other commands let users override their path to the config file, should this command also allow for that?

exp/yaml_migrate/yaml_migrate.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return err
+	}
+	logrus.Println("Successfully Opened fogg.json")
```
edulop91: probably at the logrus.Debug() level

exp/yaml_migrate/yaml_migrate.go
```diff
@@ -0,0 +1,58 @@
...
+	defer jsonFile.Close()
+
+	//Read the fogg.json file, convert it to yml format
+	byteValue, readErr := ioutil.ReadAll(jsonFile)
```
edulop91: unhandled error

apply/golden_file_test.go
```diff
@@ -11,6 +11,7 @@ import (
 
 	"github.com/chanzuckerberg/fogg/apply"
 	"github.com/chanzuckerberg/fogg/config"
+	v2 "github.com/chanzuckerberg/fogg/config/v2"
```
ryanking: same as above

exp/yaml_migrate/yaml_migrate.go
```diff
@@ -0,0 +1,58 @@
...
+	logrus.Println("Successfully created fogg.yml")
+	return nil
+}
+
+//Convert an existing json file into yml text
+func jsonToYaml(jsonFileData []byte) ([]byte, error) {
+	var jsonObj interface{}
+
+	//Convert jsonFileData into a generic interface representing json
+	err := yaml.Unmarshal(jsonFileData, &jsonObj)
```
edulop91: should this be json.Unmarshal?

apply/apply_test.go
```diff
@@ -224,49 +224,38 @@ func TestApplySmokeTest(t *testing.T) {
...
-    "staging":{
-        "components": {
-            "comp1": {},
-            "comp2": {}
-        }
-    },
-    "prod": {}
-  }
-}
+	yaml := `
```
ryanking: As discussed IRL– we should just duplicate all the test methods that have json and create yaml equivalents. Once we are ready to remove json support we can then delete the json test methods.

ryanking: Also this test is using version 1, we should only worry about yaml support for v2.

exp/yaml_migrate/yaml_migrate.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// Create YML file
+// If file does not exist one will be made, otherwise the current
+// yml file will be updated
+//TODO: See if write overwrites a yaml file or adds to yaml file
+func generateYMLFile(ymlData []byte) error {
+
+	//Write creates a new file if one does not exist
+	writeStatus := ioutil.WriteFile("fogg.yml", ymlData, 0644)
```
edulop91: by convention returned errors are named `err` or `e`

init/init.go
```diff
@@ -24,18 +23,22 @@ func userPrompt() (string, string, string, string, string, string) {
 }
 
 func writeConfig(fs afero.Fs, config *v1.Config) error {
-	json, e := json.MarshalIndent(config, "", "  ")
-	if e != nil {
-		return errs.WrapInternal(e, "unable to marshal json")
+	yaml, yamlErr := yaml.Marshal(config)
```
edulop91: same here, you can name these `yamlErr` as `err` by convention

plan/plan.go
```diff
@@ -151,13 +152,22 @@ func Eval(c *v2.Config) (*Plan, error) {
...
-		return errs.WrapInternal(err, "yaml: could not marshal")
+// Print prints a plan based on fogg file type
+func Print(p *Plan, fileName string) error {
+	switch fileName {
+	case "fogg.json":
+		out, err := json.Marshal(p)
+		if err != nil {
+			return errs.WrapInternal(err, "json: could not marshal")
+		}
+		fmt.Print(string(out))
```
edulop91: rm

plan/plan.go
```diff
@@ -151,13 +152,22 @@ func Eval(c *v2.Config) (*Plan, error) {
 	return p, nil
 }
 
-// Print prints a plan
-func Print(p *Plan) error {
-	out, err := yaml.Marshal(p)
-	if err != nil {
-		return errs.WrapInternal(err, "yaml: could not marshal")
+// Print prints a plan based on fogg file type
+func Print(p *Plan, fileName string) error {
```
edulop91: this Print is for human consumption (and not related to the config file format since it is already printing yaml). I think it would be ok to assume for now that this can print yaml files and preserve the old behavior

config/config.go
```diff
@@ -37,10 +38,21 @@ func InitConfig(project, region, bucket, table, awsProfile, owner, awsProviderVe
 
 // FindConfig loads a config and its version into memory
 func FindConfig(fs afero.Fs, configFile string) ([]byte, int, error) {
+	switch configFile {
```
ryanking: This seems like it will only read the json file if you explicitly give a `-c fogg.json`. We need to support using either yaml or json even when a config file is not specified at the command line.

Another way of thinking about this is that when we upgrade to this version, all commands should work exactly the same until you do the migration. So a repo with a json config file, when you run `fogg plan` should read the json file.

config/config.go
ryanking: This should probably just be `readConfig`.

config/config.go
ryanking: As mentioned elsewhere, we need to be a bit more nuanced in how we pick a file.

config/v1/config.go
```diff
@@ -10,63 +10,64 @@ import (
...
-	AWSRegions         []string          `json:"aws_regions,omitempty"`
-	ExtraVars          map[string]string `json:"extra_vars"`
-	InfraBucket        string            `json:"infra_s3_bucket" validate:"required"`
-	InfraDynamoTable   string            `json:"infra_dynamo_db_table"`
-	Owner              string            `json:"owner" validate:"required"`
-	Project            string            `json:"project" validate:"required"`
-	TerraformVersion   string            `json:"terraform_version" validate:"required"`
-	TfLint             *TfLint           `json:"tflint,omitempty"`
+	AccountID          int64             `json:"account_id,omitempty" yaml:"account_id,omitempty" validate:"required"`
+	AWSProfileBackend  string            `json:"aws_profile_backend" yaml:"aws_profile_backend" validate:"required"`
```
ryanking: No need to support yaml for v1

shedninja6: If I remove yaml support for v1 I fail tests in plan_test.go (Something to do with the resolvers)

apply/apply.go
```diff
@@ -12,8 +12,8 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/chanzuckerberg/fogg/config/v1"
-	"github.com/chanzuckerberg/fogg/config/v2"
+	v1 "github.com/chanzuckerberg/fogg/config/v1"
```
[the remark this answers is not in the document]
shedninja6: It doesnt change

PR #297, opened 2019-06-14
refactor: move tools to common

ryanking:
Before this tools could only be defined at the top level (not per-component).

We would like to be able to configure Atlantis separately for each component. As a bonus this will allow us to do the same for the other tools.

Note that this is a breaking change for config v2, but given that it isn't deployed widely I think we can get away with it.

### Test Plan
* unit tests

### References
*

### DIFF
--- a/config/v2/config.go
+++ b/config/v2/config.go
@@ -37,7 +37,7 @@ type Common struct {
 	Project          *string           `json:"project,omitempty" `
 	Providers        *Providers        `json:"providers,omitempty" `
 	TerraformVersion *string           `json:"terraform_version,omitempty"`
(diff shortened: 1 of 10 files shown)


config/v2/config.go
```diff
@@ -38,6 +37,7 @@ type Common struct {
 	Project          *string           `json:"project,omitempty" `
 	Providers        *Providers        `json:"providers,omitempty" `
 	TerraformVersion *string           `json:"terraform_version,omitempty"`
+	Tools            Tools             `json:"tools,omitempty"`
```
edulop91: `*Tools`? just pattern matching off the other fields here

PR #298, opened 2019-06-18
[feature] Parse config as yaml and add a command to migrate a json config to yaml

shedninja6:
#### Main Additions
- **init**: Generates .yml file by default
- **apply**: Reads .yml files (Default is .yml but also support .json)
- **plan**: Reads and writes using .yml (Default is .yml but also support .json)
- **aws-config**: Supports .yml and .json
- **tests**: All tests are configured to test both .json and .yml (excluding golden_tests) 

#### Current PR Specifics
- updated apply_test.go to support v2 yaml structure 
- Added yaml specific tests to plan_test, apply_test
- reduced json type checking code duplication in v1/config.go and v2/config.go 
- added yaml folders in testdata

#### WIP:
- Need to remove v1 and v2 inside imports [**need assistance**]
- **ISSUE**: The current yaml folders in testdata use v1 fogg structure. They are .yml files but they use v1 keys and values
- removing yaml annotations in v1/config.go causes some tests to fail (related to resolvers)


apply/apply_test.go
```diff
@@ -224,48 +225,75 @@ func TestApplySmokeTest(t *testing.T) {
...
+
+	e = Apply(fs, c, templates.Templates, false)
+	r.NoError(e)
+}
+
+func TestApplySmokeTestJSON(t *testing.T) {
+	r := require.New(t)
+	fs, _, err := util.TestFs()
+	r.NoError(err)
+	// defer os.RemoveAll(d)
```
edulop91: where did this come from?

apply/golden_file_test.go
```diff
@@ -62,20 +63,37 @@ func TestIntegration(t *testing.T) {
...
 				a.NoError(e)
-				a.NoError(afero.WriteFile(fs, "fogg.json", configContents, configMode.Mode()))
 
-				conf, e := config.FindAndReadConfig(fs, "fogg.json")
-				a.NoError(e)
-				fmt.Printf("conf %#v\n", conf)
+				var configMode os.FileInfo
+				var conf *v2.Config
+				if isYaml {
+					configMode, e = testdataFs.Stat("fogg.yml")
```
edulop91: the only thing that changes between this conditional and the other is `fogg.{json|yml}`, maybe you can do something like:
```
filename := "fogg.yml"
if !isYaml {
filename = "fogg.json"
}

// So all of this code is not duplicated anymore
configMode, e = testdataFs.Stat(filename)
....

```

cmd/upgrade.go
```diff
@@ -10,7 +10,8 @@ import (
 )
 
 func init() {
-	upgrade.Flags().StringP("config", "c", "fogg.json", "Use this to override the fogg config file.")
+	//Defaults to fogg.yml but fogg.json can be used as an argument instead
```
edulop91: i feel like this comment is implied, so there might not be much value in keeping it

config/config.go
```diff
@@ -41,10 +41,8 @@ func FindConfig(fs afero.Fs, configFile string) ([]byte, int, error) {
 	if err != nil {
 		return nil, 0, errs.WrapUser(err, "unable to open config file")
 	}
-	reader := io.ReadCloser(f)
-	defer reader.Close()
 
-	b, e := ioutil.ReadAll(reader)
+	b, e := ioutil.ReadAll(f)
```
edulop91: you might need to `defer f.Close()` still

config/config.go
```diff
@@ -62,23 +60,28 @@ func FindAndReadConfig(fs afero.Fs, configFile string) (*v2.Config, error) {
...
+
 		c, err := v1.ReadConfig(b)
 		if err != nil {
 			return nil, err
 		}
 
-		c2, err := UpgradeConfigVersion(c)
-		return c2, err
-
+		c2, err = UpgradeConfigVersion(c)
```
edulop91: why change this code?

config/v2/config.go
```diff
@@ -8,101 +8,112 @@ import (
...
 	}
-	e := json.Unmarshal(b, c)
 
-	return c, errs.WrapUser(e, "unable to parse config")
+	if v1.IsJSON(b) {
+		e = json.Unmarshal(b, c)
+	} else {
+		e = yaml.Unmarshal(b, c)
+	}
+	if e != nil {
```
edulop91: can probably do without this conditional and just `return c, errs.WrapUser(e, ...)`

exp/yaml_migrate/yaml_migrate.go
```diff
@@ -0,0 +1,70 @@
...
+	return yaml.Marshal(jsonObj)
+}
+
+// Create YML file
+// If file does not exist one will be made, otherwise the current
+// yml file will be updated
+func generateYMLFile(ymlData []byte) error {
+
+	//Write creates a new file if one does not exist
+	err := ioutil.WriteFile("fogg.yml", ymlData, 0644)
```
edulop91: hmm wondering if we should allow users to specify the `fogg.yml` or output file

init/init.go
```diff
@@ -24,18 +23,22 @@ func userPrompt() (string, string, string, string, string, string) {
 }
 
 func writeConfig(fs afero.Fs, config *v1.Config) error {
-	json, e := json.MarshalIndent(config, "", "  ")
-	if e != nil {
-		return errs.WrapInternal(e, "unable to marshal json")
+	yaml, err := yaml.Marshal(config)
+
```
edulop91: rm extra whitespace

init/init.go
```diff
@@ -24,18 +23,22 @@ func userPrompt() (string, string, string, string, string, string) {
...
-		return errs.WrapInternal(e, "unable to create config file fogg.json")
+
+	yamlConfigFile, err := fs.Create("fogg.yml")
+	if err != nil {
+		return errs.WrapInternal(err, "unable to create config file fogg.yml")
 	}
-	_, e3 := configFile.Write(json)
-	return e3
+	_, err = yamlConfigFile.Write(yaml)
+
```
edulop91: rm extra whitespace

plan/plan.go
```diff
@@ -151,11 +151,11 @@ func Eval(c *v2.Config) (*Plan, error) {
...
 }
 
-// Print prints a plan
-func Print(p *Plan) error {
+// Print prints a plan based on fogg file type
+func Print(p *Plan, fileName string) error {
 	out, err := yaml.Marshal(p)
 	if err != nil {
-		return errs.WrapInternal(err, "yaml: could not marshal")
+		return errs.WrapInternal(err, "json: could not marshal")
```
edulop91: ~json~yaml

apply/apply_test.go
shedninja6: This was in the original codebase, but I can remove it

config/config.go
shedninja6: I changed it to avoid duplication. The return statement occurred twice. Is that something I should consider changing?

exp/yaml_migrate/yaml_migrate.go
shedninja6: I tried that earlier but since c is initialized to a Config, the change fails all of the nil tests.

config/v2/config.go
shedninja6: I tried that earlier but since c is initialized to a Config, the change fails all of the nil tests.

init/init.go
shedninja6: Does the team have a set of general best practices for go or just in general?

exp/yaml_migrate/yaml_migrate.go
shedninja6: Not sure what you mean? Are you saying we should give users the ability to name the file?

edulop91: let's keep as is I think that's better for now

apply/apply_test.go
edulop91: thanks, I think keeping it there is confusing since it doesn't look like we even have `d` variable in scope

apply/golden_file_test.go
```diff
@@ -60,18 +61,26 @@ func TestIntegration(t *testing.T) {
...
-				configContents, e := afero.ReadFile(testdataFs, "fogg.json")
+				configContents, e := afero.ReadFile(testdataFs, "fogg.yml")
+				if os.IsNotExist(e) { //If the error is related to the file being non-existent
+					configContents, e = afero.ReadFile(testdataFs, "fogg.json")
+					fileName = "fogg.json"
+				}
 				a.NoError(e)
-				configMode, e := testdataFs.Stat("fogg.json")
+
+				var configMode os.FileInfo
```
edulop91: you can now get rid of these `var` declarations and do the inline assignments on L79,83

apply/golden_file_test.go
```diff
@@ -60,18 +61,26 @@ func TestIntegration(t *testing.T) {
...
 			} else {
-
+				fileName := "fogg.yml"
 				fs, _, e := util.TestFs()
 				a.NoError(e)
 
 				// copy fogg.json into the tmp test dir (so that it doesn't show up as a diff)
-				configContents, e := afero.ReadFile(testdataFs, "fogg.json")
+				configContents, e := afero.ReadFile(testdataFs, "fogg.yml")
+				if os.IsNotExist(e) { //If the error is related to the file being non-existent
```
edulop91: nice :ok_hand:

config/config.go
```diff
@@ -41,10 +41,10 @@ func FindConfig(fs afero.Fs, configFile string) ([]byte, int, error) {
 	if err != nil {
 		return nil, 0, errs.WrapUser(err, "unable to open config file")
 	}
-	reader := io.ReadCloser(f)
-	defer reader.Close()
 
-	b, e := ioutil.ReadAll(reader)
+	b, e := ioutil.ReadAll(f)
+	defer f.Close()
```
edulop91: thanks! but maybe move it right after line 43 so it is next to the place where you create the file (and it is easier for readers to keep track of)

config/config.go
edulop91: [this is just a style thing so opinions vary] I think that adding distance in the file between where the error/value is being generated and when it is being used adds for more complexity/makes it harder to read. 
I actually think the better solution would be to just `return UpgradeConfigVersion` or `return v2.ReadConfig` directly

init/init.go
edulop91: we don't yet (and this is just a matter of style so opinions differ)

plan/plan.go
```diff
@@ -142,8 +142,8 @@ func Eval(c *v2.Config) (*Plan, error) {
 	return p, nil
 }
 
-// Print prints a plan
-func Print(p *Plan) error {
+// Print prints a plan based on fogg file type
+func Print(p *Plan, fileName string) error {
```
edulop91: looks like `fileName` is unused

cmd/plan.go
```diff
@@ -52,6 +52,7 @@ var planCmd = &cobra.Command{
 			return e
 		}
 
+		//Pass configFile name to print using the correct format
```
edulop91: comment not relevant anymore

config/config.go
edulop91: ping

config/v1/config.go
```diff
@@ -91,65 +95,80 @@ const (
...
 	}
-	e := json.Unmarshal(b, c)
+
+	if IsJSON(b) {
+		e = json.Unmarshal(b, c)
+		logrus.Warn("JSON is deprecated, consider migrating to yaml")
+	} else {
+		e = yaml.Unmarshal(b, c)
+	}
 	if e != nil {
```
edulop91: looks like you can get rid of this branch since you're already returning https://github.com/chanzuckerberg/fogg/pull/298/files#diff-374f4b941ce2213d7811071542d264a6R164

exp/yaml_migrate/yaml_migrate.go
edulop91: just `return ioutil.WriteFile(...)` since you're not doing anything with that error (other than returning it)

config/v1/config.go
```diff
@@ -1,72 +1,76 @@
...
+	InfraBucket        string            `json:"infra_s3_bucket" yaml:"infra_s3_bucket" validate:"required"`
+	InfraDynamoTable   string            `json:"infra_dynamo_db_table" yaml:"infra_dynamo_db_table" `
+	Owner              string            `json:"owner" yaml:"owner" validate:"required"`
+	Project            string            `json:"project" yaml:"project" validate:"required"`
+	TerraformVersion   string            `json:"terraform_version" yaml:"terraform_version" validate:"required"`
+	TfLint             *TfLint           `json:"tflint,omitempty" yaml:"tflint,omitempty"`
 }
 
 type Account struct {
-	AccountID          *json.Number      `json:"account_id"`
```
edulop91: this shuold still be `json.Number`

config/v1/config.go
shedninja6: Removing it fails the nil tests because c is still initialized.

edulop91:
@shedninja6 there are a handful of merge conflicts, can you fix those and then I'll review?

apply/apply_test.go
```diff
@@ -222,52 +224,80 @@ func TestApplySmokeTest(t *testing.T) {
...
+		"aws_profile_provider": "prof",
+		"aws_profile_backend": "prof",
+		"aws_provider_version": "0.12.0",
+		"account_id": 789,
+		"infra_s3_bucket": "buck",
+		"project": "proj",
+		"terraform_version": "0.100.0",
+		"owner": "foo@example.com"
+	  },
+	  "travis_ci": {
```
edulop91: looks like a bad merge here

cmd/yaml.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/chanzuckerberg/fogg/util"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	rootCmd.AddCommand(yamlMigrateCmd)
+}
+
+var yamlMigrateCmd = &cobra.Command{
```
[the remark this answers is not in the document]
edulop91: deferring to a followup pr

config/config.go
```diff
@@ -62,33 +63,49 @@ func FindAndReadConfig(fs afero.Fs, configFile string) (*v2.Config, error) {
...
 }
 
 type ver struct {
-	Version int `json:"version"`
+	Version int `json:"version" yaml:"version"`
 }
 
 // detectVersion will detect the version of a config
-func detectVersion(b []byte) (int, error) {
+func detectVersion(b []byte, fs afero.Fs, configFile string) (int, error) {
```
edulop91: by convention, `fs` is the first argument to functions

config/config.go
```diff
@@ -62,33 +63,49 @@ func FindAndReadConfig(fs afero.Fs, configFile string) (*v2.Config, error) {
...
 	v := &ver{}
-	err := json.Unmarshal(b, v)
+	var err error
+
+	info, err := fs.Stat(configFile)
+	if err != nil {
+		return 0, errs.WrapUser(err, "unable to find file")
```
edulop91: this error could be more than just unable to find that file

edulop91:
```suggestion
		return 0, errs.WrapUser(err, "unable to stat %s", configFile)
```

config/upgrade.go
```diff
@@ -26,9 +25,10 @@ func Upgrade(fs afero.Fs, configFile string) error {
 			return err
 		}
 
-		marshalled, err := json.MarshalIndent(c2, "", "  ")
+		//Should this marshal it to v2?
```
edulop91: what's this comment?

config/v1/config.go
```diff
@@ -143,13 +143,16 @@ type Config struct {
 }
 
 func ReadConfig(b []byte) (*Config, error) {
+	var e error
```
edulop91:
```suggestion
```

config/v1/config.go
```diff
@@ -143,13 +143,16 @@ type Config struct {
...
 	c := &Config{
 		Docker: true,
 	}
-	e := json.Unmarshal(b, c)
+
+	e = json.Unmarshal(b, c)
```
edulop91:
```suggestion
	e := json.Unmarshal(b, c)
```

config/v2/config.go
```diff
@@ -4,54 +4,72 @@ import (
...
 
 	v1 "github.com/chanzuckerberg/fogg/config/v1"
 	"github.com/chanzuckerberg/fogg/errs"
+	"github.com/spf13/afero"
+	"gopkg.in/yaml.v2"
 )
 
-func ReadConfig(b []byte) (*Config, error) {
+//ReadConfig take a byte array as input and outputs a json or yaml config file
+func ReadConfig(b []byte, fs afero.Fs, configFile string) (*Config, error) {
```
edulop91: we've been saying `fs` is the first argument by convention

config/v2/config.go
```diff
@@ -4,54 +4,72 @@ import (
...
+	var e error
```
edulop91: confusing to have both an `e` and an `err` in scope, maybe standardize on one

config/config.go
```diff
@@ -62,33 +63,49 @@ func FindAndReadConfig(fs afero.Fs, configFile string) (*v2.Config, error) {
...
+	}
+
+	//Unmarshals based on file extension
+	switch filepath.Ext(info.Name()) {
+	case ".yml":
```
edulop91:
```suggestion
	case ".yml", ".yaml":
```
Both are [common](https://yaml.org/faq.html) extensions

migrate/yaml.go
```diff
@@ -0,0 +1,23 @@
...
+	config, err := config.FindAndReadConfig(fs, configFile)
+	if err != nil {
+		return err
+	}
+
+	yml, err := yaml.Marshal(config)
+	if err != nil {
+		return err
+	}
+	afero.WriteFile(fs, "fogg.yml", yml, 0644)
```
edulop91: return error being ignored

config/upgrade.go
shedninja6: That was a mental note that I forgot to remove

apply/apply_test.go
shedninja6: Not sure what you're referring to

config/v1/config.go
```diff
@@ -23,10 +23,10 @@ type Defaults struct {
 	AWSProviderVersion string            `json:"aws_provider_version" validate:"required"`
 	AWSRegionBackend   string            `json:"aws_region_backend" validate:"required"`
 	AWSRegionProvider  string            `json:"aws_region_provider" validate:"required"`
-	AWSRegions         []string          `json:"aws_regions,omitempty"`
-	ExtraVars          map[string]string `json:"extra_vars"`
+	AWSRegions         []string          `json:"aws_regions,omitempty" `
```
ryanking: This extra whitespace should probably not be added.

config/v1/config_test.go
```diff
@@ -84,8 +95,8 @@ func TestParse(t *testing.T) {
 }
 
 func TestJsonFailure(t *testing.T) {
```
ryanking:
```suggestion
func TestYamlFailure(t *testing.T) {
```

config/v2/config.go
```diff
@@ -4,54 +4,72 @@ import (
...
 	"reflect"
 
 	v1 "github.com/chanzuckerberg/fogg/config/v1"
 	"github.com/chanzuckerberg/fogg/errs"
+	"github.com/spf13/afero"
+	"gopkg.in/yaml.v2"
 )
 
-func ReadConfig(b []byte) (*Config, error) {
+//ReadConfig take a byte array as input and outputs a json or yaml config file
```
ryanking:
```suggestion
//ReadConfig take a byte array as input and outputs a json or yaml config struct
```

config/v2/config.go
```diff
@@ -4,54 +4,72 @@ import (
...
 	}
-	e := json.Unmarshal(b, c)
 
-	return c, errs.WrapUser(e, "unable to parse config")
+	info, e := fs.Stat(configFile)
+	if e != nil {
+		return nil, errs.WrapUser(e, "unable to find file")
+	}
+
+	//Determines the extension of the file
```
ryanking:
```suggestion
//Determines the file type based on the extension of the file
```

config/v2/config.go
```diff
@@ -4,54 +4,72 @@ import (
...
+	switch filepath.Ext(info.Name()) {
+	case ".yml", ".yaml":
+		e = yaml.Unmarshal(b, c)
+	case ".json":
+		e = json.Unmarshal(b, c)
+	default:
+		return nil, errs.NewUser("File type is not supported")
```
ryanking:
```suggestion
return nil, errs.NewUserf("File type %s is not supported", filepath.Ext(info.Name()))
```

... but refactor so that you don't have to run filepath.Ext(info.Name()) twice

init/init.go
```diff
@@ -24,18 +23,20 @@ func userPrompt() (string, string, string, string, string, string) {
 }
 
 func writeConfig(fs afero.Fs, config *v1.Config) error {
```
ryanking: just a reminder that this should be `v2`

plan/plan.go
```diff
@@ -29,16 +29,17 @@ type Common struct {
 	TerraformVersion string `yaml:"terraform_version"`
 }
 
+//ComponentCommon represents common componenet fields
```
ryanking:
```suggestion
//ComponentCommon represents common fields for components
```

apply/apply_test.go
edulop91: defaults.tools.travis_ci and then on the right travis_ci is not nested under tools and not under defaults

edulop91: is that intentional?

edulop91: I assumed it might've been the result of a merge conflict since the format on the right is not a valid v2 format anymore

plan/plan.go
```diff
@@ -168,7 +173,7 @@ func Eval(c *v2.Config) (*Plan, error) {
 	return p, nil
 }
 
-// Print prints a plan
+// Print prints a plan based on fogg file type
```
edulop91:
```suggestion
// Print prints a plan
```

config/config.go
```diff
@@ -2,35 +2,44 @@ package config
...
-			AWSProviderVersion: awsProviderVersion,
-			AWSRegionBackend:   region,
-			AWSRegionProvider:  region,
-			ExtraVars:          map[string]string{},
-			InfraBucket:        bucket,
-			InfraDynamoTable:   table,
-			Owner:              owner,
-			Project:            project,
-			TerraformVersion:   "0.11.7",
+//InitConfig InitConfig initializes the config file using user input
```
ryanking:
```suggestion
//InitConfig initializes the config file using user input

```

PR #303, opened 2019-06-20
[feature] atlantis integration

### LINKED ISSUE #278: atlantis support

### LINKED ISSUE #289: fogg generate aws/config for atlantis

### LINKED ISSUE #290: make targets/mode for Atlantis operations

ryanking:
This PR adds Atlantis (<runatlantis.io>) integration to fogg. The goal is to be able to use fogg to generate an atlantis.yaml configuration file and all related tools, scripts etc.

Happy to take feedback on both the approach and the implementation.

### Approach

The basic approach is, in no particular order–

* To the degree possible it would be great to be able to make changes to how Atlantis works without touching the Atlantis instances, so as much as possible is put into the per-repo config file.
* We allow configuring Atlantis on a per-component basis. I think this is important as I would expect teams to adopt it incrementally.
* We do a 2-phase plan - first . plan for each component (so that we can set a var in each) and then walk those plans and generate a plan for the atlantis.yaml and aws config
* We generate an aws config for atlantis. Because we rely on aws profiles for authentication, we need to find a way to get the Atlantis instance configured with all the right profiles. In the future we may want to think about doing this for other modes.
* If atlantis is enabled for a component, `make apply` will by default fail. You have to add `FORCE=1` to override. This should nudge people in the right direction while allowing a fall-back to running locally (important for adoption phase).
* we add a `make setup` that will install fogg and run `fogg setup` (you can run this from components, not just top-level)
* we no longer run `terraform get` since that is redundant with `terrraform init` and we **don't run -update=true anymore** which means that modules will not get updated. People should be using stable versions anyway and this makes it much faster
* tfenv is now installed inside the repo, rather than homedir. This makes it so that we can use tfenv in atlantis (rather than atlantis' own process for getting terraform versions)

--- 

Note there are also a handful of stylistic changes in this diff. Tried to keep those to a minimum, but also wanted to make some improvements, especially to our makefiles.

## Test Plan
* unit tests
* ran on shared-infra repo for 1 component

## References
* <https://www.runatlantis.io/docs/>
* Fixes #290 
* Fixes #289
* Fixes #278


Makefile
```diff
@@ -18,10 +18,14 @@ lint: ## run the fast go linters
 	golangci-lint run
 .PHONY: lint
 
-packr: ## run the packr tool to generate our static files
+TEMPLATES := $(shell find templates -not -name "*.go")
+
+templates/a_templates-packr.go: $(TEMPLATES)
```
ryanking: This *should* make it so that packr is only run if there are changed templates.

config/v2/config.go
```diff
@@ -107,6 +114,7 @@ type SnowflakeProvider struct {
 }
 
 type Backend struct {
+	AccountID   *string `json:"account_id,omitempty"`
```
ryanking: we need this to generate the aws config for atlantis

config/v2/validation.go
```diff
@@ -173,6 +170,27 @@ func (c *Config) ValidateOktaProviders() error {
...
 }
 
+func (c *Config) ValidateAtlantis() error {
+	var errs *multierror.Error
+	c.WalkComponents(func(component string, comms ...Common) {
+		a := ResolveAtlantis(comms...)
+		// TODO refactor to a ResolveBackend method
+
+		// a can no longer be nil
+		if a.Enabled != nil && *a.Enabled {
```
edulop91: should we ditch enabled and instead go off the presence/absence or roleName?

config/v2/resolvers.go
```diff
@@ -175,6 +176,24 @@ func ResolveTfLint(commons ...Common) v1.TfLint {
 	}
 }
 
+func ResolveAtlantis(commons ...Common) Atlantis {
```
edulop91: should it be `*Atlantis`? pattern matching against the other resolvers

config/v2/resolvers.go
```diff
@@ -175,6 +176,24 @@ func ResolveTfLint(commons ...Common) v1.TfLint {
...
+		if c.Tools != nil && c.Tools.Atlantis != nil && c.Tools.Atlantis.Enabled != nil {
+			enabled = *c.Tools.Atlantis.Enabled
+		}
+	}
+
+	roleName := lastNonNil(AtlantisRoleNameGetter, commons...)
+	rolePath := lastNonNil(AtlantisRolePathGetter, commons...)
+
+	return Atlantis{
+		Enabled:  &enabled,
```
edulop91: should we ditch Enabled and instead go off the presence/absence of the atlantis block?

templates/common/fogg_header.tmpl
```diff
@@ -0,0 +1,4 @@
+{{define "fogg_header"}}
```
ryanking: todo later, use this everywhere

plan/atlantis.go
```diff
@@ -0,0 +1,131 @@
...
+	"fmt"
+	"sort"
+)
+
+type Atlantis struct {
+	Enabled  bool
+	Projects []AtlantisProject
+
+	// Profiles is a map of profile name -> role infro
+	Profiles map[string]AWSRole
```
edulop91: AWSProfiles maybe? I assume we might have to do similar things for other types of providers, should this be
```
Providers {
AWS: {}
}
```
to mirror how the v2 config specifies providers?

config/v2/resolvers.go
edulop91: oh, I see. you need Enabled to properly resolve the inheritance chain

config/v2/resolvers.go
ryanking: yeah, good catch

config/v2/resolvers.go
ryanking: I don't think that would work in the case that you turn it on in defaults and want to turn it off somewhere else.

config/v2/validation.go
```diff
@@ -64,10 +65,6 @@ func (c *Config) Validate() ([]string, error) {
 	return warnings, errs.ErrorOrNil()
 }
 
-func merge(warnings []string, err *multierror.Error, w []string, e error) ([]string, *multierror.Error) {
```
ryanking: was not in use

templates/repo/scripts/failed_output_only
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
ryanking: I was trying to get this to work to suppress non-error output when running in atlantis. I didn't quite get it working, but going to leave it here to iterate on later.

templates/repo/scripts/component.mk
```diff
@@ -8,41 +8,71 @@ include $(SELF_DIR)/common.mk
 all:
 .PHONY: all
 
+setup:
+	$(MAKE) -C $(REPO_ROOT) setup
```
edulop91: .PHONY

templates/repo/scripts/component.mk
```diff
@@ -8,41 +8,71 @@ include $(SELF_DIR)/common.mk
...
-.PHONY: get
-
-plan: terraform fmt get init
-	$(terraform_command) plan
+ifeq ($(MODE),local)
+plan: init fmt
+	@$(terraform_command) plan
+else ifeq ($(MODE),atlantis)
+plan: init lint
+	@$(terraform_command) plan -input=false -no-color -out $(PLANFILE) | scenery
```
edulop91: what's `scenery`?

templates/repo/scripts/failed_output_only
edulop91: maybe leave a TODO on the file for future selves

testdata/bless_provider/scripts/component.mk
```diff
@@ -8,41 +8,71 @@ include $(SELF_DIR)/common.mk
 all:
 .PHONY: all
 
+setup:
+	$(MAKE) -C $(REPO_ROOT) setup
```
edulop91: does this need a .PHONY?

templates/repo/scripts/component.mk
ryanking: A terraform output prettifier written in go https://github.com/dmlittle/scenery

I have it baked into the atlantis image. Considering using fogg to install it for local mode too, but I want to see what happens with 0.12 output (it is supposedly better).

testdata/bless_provider/scripts/component.mk
ryanking: yeah, will get generated by fixiing above

PR #309, opened 2019-06-26
[feature] Move terraform validate from lint stage to plan stage

edulop91:
### Summary
Having `terraform validate` as a substep of `lint` forces us to have fully configured providers (including credentials/secrets). I would like to be able to lint code without requiring these secrets be present. One way could be moving this validation over to the plan stage (although this feels somewhat redundant).  

### References

### DIFF
--- a/templates/repo/scripts/component.mk
+++ b/templates/repo/scripts/component.mk
@@ -21,7 +21,7 @@ fmt: terraform
 	@echo
 .PHONY: fmt
 
-lint: terraform-validate lint-terraform-fmt lint-tflint
+lint: lint-terraform-fmt lint-tflint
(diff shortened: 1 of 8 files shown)


ryanking:
> This doesn't buy us anything over running plan. I say remove it from lint but don't add it to plan.

I am doing to dismiss my review so you don't have to block on me, assuming you are going to make this change.

templates/repo/scripts/component.mk
```diff
@@ -33,10 +33,6 @@ else
 endif
 .PHONY: lint-tflint
 
-terraform-validate: terraform init
```
edulop91: unused so I just got rid of it

PR #312, opened 2019-06-27
[fix] buildevents is now published at a different location

edulop91:
### Summary
https://github.com/honeycombio/buildevents/pull/24/files was a breaking change to how buildevents artifacts are released

### Test Plan
unittests

### References


edulop91:
TBR @chanzuckerberg/czi-shared-infra

PR #313, opened 2019-06-27
[fix] Fix missing terraform_version tag and upgrade yaml dep

edulop91:
### Summary
We were missing a yaml tag in v1.modules for terraform_version which caused yaml conversion to look like terraformversion. Also upgrading the yaml package version to latest

### Test Plan
unittests


edulop91:
TBR @chanzuckerberg/czi-shared-infra

PR #319, opened 2019-07-17
[fix] Added clearer logging and un-nested errors

shedninja6:
### Summary
A user ran into a situation where he could not debug his issues because the migration logging was not clear. I added logs to help users debug configuration file errors. I also made existing errors more clear by un-nesting errors.

### Test Plan
Fixes did not effect tests 

### References


config/config.go
```diff
@@ -50,7 +50,8 @@ func InitConfig(project, region, bucket, table, awsProfile, owner, awsProviderVe
 func FindConfig(fs afero.Fs, configFile string) ([]byte, int, error) {
 	f, err := fs.Open(configFile)
 	if err != nil {
-		return nil, 0, errs.WrapUserf(err, "unable to open config file %s (Maybe specify the config file with a --config flag)", configFile)
+		logrus.Info("Fogg configuration files were recently migrated to .yml, run 'fogg migrate' to update Fogg")
```
edulop91: I think we should go a bit further here and check if that error is a `os.IsNotExist(err)` (and perhaps even check if `fogg.json` is present. So we can give a better error for users. 

I also think we should add a comment saying to remove these checks. Somethign like
```
 // TODO(ec): remove this deprecation warning
```

Issue #310, opened 2019-06-26
turn off refresh


ryanking:
We should have a config option to turn off refresh (by passing -refresh=false) to plan and apply.

This greatly speeds up operations at the risk of missing some drift.

PR #321, opened 2019-07-17
[feature] Optionally turn off refresh for plan/apply using REFRESH_ENABLED

edulop91:
### Summary
Fixes https://github.com/chanzuckerberg/fogg/issues/310. This turns off refresh for plan/apply. This gives us quicker plan/apply at the expense of some correctness checks.  

### Test Plan
unittests + apply in repos

### References
- https://www.terraform.io/docs/commands/refresh.html
- https://www.terraform.io/docs/commands/plan.html#refresh-true

### DIFF
--- a/templates/repo/scripts/common.mk
+++ b/templates/repo/scripts/common.mk
@@ -6,6 +6,8 @@ TF_VARS := $(patsubst %,-e%,$(filter TF_VAR_%,$(.VARIABLES)))
 REPO_ROOT := $(shell git rev-parse --show-toplevel)
 REPO_RELATIVE_PATH := $(shell git rev-parse --show-prefix)
 AUTO_APPROVE := false
+REFRESH_ENABLED := true # Should terraform refresh during plan/apply
+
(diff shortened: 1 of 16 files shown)


chauvm:
Should we make a flag for refresh?

chauvm:
I guess I don't understand why we don't want to ever refresh before applying changes. We may want to skip refresh most of the time when developing something, but should not always do so, right?

edulop91:
>I guess I don't understand why we don't want to ever refresh before applying changes. We may want to skip refresh most of the time when developing something, but should not always do so, right?

Maybe. If everything is managed through terraform then feels like refresh is wasted work. If we enforce that invariant then we'd never really want to refresh. Also requires people to use immutable references to modules (using `master` would mean that local copies of the modules do not get refreshed). I agree with adding a flag to control this though.

edulop91:
The speedup is quite significant which makes me think we want this.

chauvm:
Sounds good!

PR #322, opened 2019-07-19
[feature] Cache fogg setup -- custom plugins and providers

edulop91:
### Summary
This will cache `fogg setup` -- custom plugins and providers to the fogg cache (currently at `~/.fogg/cache/custom_plugins`)

### Test Plan
unittests

### DIFF
--- a/plugins/custom_plugin.go
+++ b/plugins/custom_plugin.go
@@ -5,9 +5,9 @@ import (
 	"archive/zip"
 	"bytes"
 	"compress/gzip"
-	"fmt"
+	"crypto/sha256"
+	"encoding/base64"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"path"
@@ -17,6 +17,8 @@ import (
 	"text/template"
 
 	"github.com/chanzuckerberg/fogg/errs"
+	"github.com/chanzuckerberg/fogg/util"
+	"github.com/peterbourgon/diskv"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/afero"
 )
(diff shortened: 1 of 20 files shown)


plugins/custom_plugin.go
```diff
@@ -73,18 +80,20 @@ func (cp *CustomPlugin) install(
 		return errs.NewUser("nil fs")
 	}
 
-	fullUrl, err := Template(url, pluginOS, pluginArch)
+	fullURL, err := Template(url, pluginOS, pluginArch)
 	if err != nil {
+		logrus.Error(err)
```
edulop91:
```suggestion
```

plugins/custom_plugin.go
```diff
@@ -151,43 +176,32 @@ func (cp *CustomPlugin) processBin(fs afero.Fs, name string, downloadPath string
...
 	dst, err := fs.Create(targetPath)
 	if err != nil {
 		return errs.WrapUserf(err, "Could not open target file at %s", targetPath)
 	}
 
-	_, err = io.Copy(dst, src)
+	_, err = io.Copy(dst, file)
 	if err != nil {
-		return errs.WrapUserf(err, "Could not move %s to %s", downloadPath, targetPath)
+		return errs.WrapUserf(err, "Could not move read to %s", targetPath)
```
edulop91:
```suggestion
		return errs.WrapUserf(err, "Could not move plugin to %s", targetPath)
```

util/template_test.go
```diff
@@ -3,56 +3,14 @@ package util
 import (
 	"testing"
 
-	"github.com/chanzuckerberg/fogg/templates"
```
edulop91: this was causing cirular dependencies so moved it to templates package

PR #326, opened 2019-07-22
[fix] Enable fogg.tf remote state data source for accounts without an aws provider

edulop91:
### Test Plan
unittests

### DIFF
--- a/plan/plan.go
+++ b/plan/plan.go
@@ -114,8 +114,8 @@ type Module struct {
 type Account struct {
 	ComponentCommon `json:",inline" yaml:",inline"`
 
-	AllAccounts map[string]json.Number `yaml:"all_accounts"`
-	AccountName string                 `yaml:"account_name"`
+	AllAccounts map[string]*json.Number `yaml:"all_accounts"`
+	AccountName string                  `yaml:"account_name"`
 	Global      *Component
 }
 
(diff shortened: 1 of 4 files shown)


plan/plan.go
```diff
@@ -408,12 +407,14 @@ func resolveExtraVars(vars ...map[string]string) map[string]string {
...
 
-func resolveAccounts(accounts map[string]v2.Account) map[string]json.Number {
-	a := make(map[string]json.Number)
+func resolveAccounts(accounts map[string]v2.Account) map[string]*json.Number {
+	a := make(map[string]*json.Number)
 	for name, account := range accounts {
-		if account.Providers != nil && account.Providers.AWS != nil && account.Providers.AWS.AccountID != nil {
-			a[name] = *account.Providers.AWS.AccountID
+		var acctID *json.Number
+		if account.Providers != nil && account.Providers.AWS != nil {
```
shedninja6: Just curious, how come you removed the`account.Providers.AWS.AccountID != nil` conditional?

edulop91: it's ok for it to be nil, we then do a corresponding [nil check](https://github.com/chanzuckerberg/fogg/blob/master/templates/account/fogg.tf.tmpl#L79) in the templates (which we were already doing)

PR #327, opened 2019-07-22
[fix] Allow env overwrite of ENABLE_REFRESH

edulop91:
### Test Plan
unittests


chauvm:
hmm it's weird that AUTO_APPROVE flag did not require this and still can be overridden.

edulop91:
@chauvm I wonder if it is a difference between BSD/GNU make or something

PR #329, opened 2019-07-26
[feature] Feature to detect terraform file version changes

shedninja6:
### Summary
Created a command that will search the terraform files of a Fogg project to see if they are using the latest module versions. **Currently:** The command, given a path, will successfully traverse the project to gather all of the terraform modules that are being used. Using that information, it loads the latest version of each module from GitHub or the terraform registry. **Comparisons have not yet been implemented** 

The main files are located in `./cmd/exp/examine` and `./exp/examine`

### Test Plan
Unit tests and testdata were created to make sure that functionality is working; however, the tests are not extensive.

### References
- [Using Tags](https://github.com/src-d/go-git/issues/430)
- [Downloading external/github modules](https://github.com/hashicorp/go-getter)
- [Loading local terraform modules](https://godoc.org/github.com/hashicorp/terraform-config-inspect/tfconfig)
- [CLI to test reading modules](https://github.com/hashicorp/terraform-config-inspect)


exp/examine/local.go
```diff
@@ -0,0 +1,164 @@
...
+
+	//BFS on the current queue of submodules to find their submodules
+	for { //Do: recursively add new submodules to queue, While: queue is not empty
+
+		submods, err := getSubmodules(fs, queue[0].module.Path)
+		if err == nil {
+			//Add new elements to temp
+			queue = append(queue, submods...)
+			submodules = append(submodules, submods...)
+		}
```
redhoteggplant: would it be useful to log a message or accumulate errors if `getSubmodules()` returns an error at any point?

exp/examine/local.go
```diff
@@ -0,0 +1,164 @@
...
+
+	//Find all submodules for the current directory
+	submodules, err := getSubmodules(fs, dir)
+	if err != nil {
+		return nil, err
+	}
+	queue = append(queue, submodules...)
+
+	//BFS on the current queue of submodules to find their submodules
+	for { //Do: recursively add new submodules to queue, While: queue is not empty
```
redhoteggplant:
```suggestion
	for len(queue) > 0 { //Do: recursively add new submodules to queue, While: queue is not empty
```

redhoteggplant: by specifying this loop condition we can remove lines 57-59 for cleaner code!

exp/examine/local.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	//FIXME: Make contains check module path so that repeats are not added
+	// if(modules != nil && contains(modules, dir))
+	var modules []ModuleWrapper
+	mod, diag := tfconfig.LoadModule(dir)
+	if diag.HasErrors() {
+		return nil, errs.WrapInternal(diag.Err(), "There was an issue loading the module")
```
redhoteggplant: just curious, why wrap some errors and not others?

exp/examine/local.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	//Loads all modules that the current module depends on
+	sources := getSources(mod)
+	for i, source := range sources {
+		if strings.HasPrefix(source.moduleSource, githubURL) { // If the module is a github link, specifically cztack
```
redhoteggplant: oo how do you know it's from cztack?

exp/examine/util.go
```diff
@@ -0,0 +1,103 @@
...
+	//Load module into struct
+	mod, diag := tfconfig.LoadModule(tmpDir)
+	if diag.HasErrors() {
+		return nil, errs.WrapInternal(diag.Err(), "There was an issue loading the module")
+	}
+
+	return mod, nil
+}
+
+func openGitOrExit(fs afero.Fs) {
```
redhoteggplant: is this func used anywhere?

exp/examine/local.go
```diff
@@ -0,0 +1,164 @@
...
+
+//GetLocalModules retrieves all terraform modules within a given directory
+//TODO:(EC) Define local and global modules OR rename the values
+func GetLocalModules(fs afero.Fs, dir string) ([]ModuleWrapper, error) {
+	_, err := os.Stat(dir)
+	if err != nil {
+		return nil, err
+	}
+
+	modules, err := getAllModules(fs, dir)
```
redhoteggplant:
```suggestion
	return getAllModules(fs, dir)
```
and then delete lines 28-31

exp/examine/local.go
shedninja6: Good point!

exp/examine/local.go
shedninja6: I wrapped the innermost errors. If I wrap all of the errors the messages wont be super clear. 
They would look something like this:
`There was an issue loading the module: There was an issue loading the module: There was an issue loading the module: ...`

exp/examine/local.go
shedninja6: The files that I have been testing on have modules that are from cztack. I plan on making this more general, so that all modules (that are from github) can be parsed.

exp/examine/util.go
shedninja6: Nope, good catch!

exp/examine/latest.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+	}
+
+	return latestModules, nil
+}
+
+//generateURL creates github url for the given module
+func generateURL(fs afero.Fs, module ModuleWrapper) (string, error) {
+	url := ""
+	if strings.HasPrefix(module.moduleSource, "github.com/chanzuckerberg") { //Clone the repo and get the latest tag
```
ryanking: Don't clone the repo - just use the github api – https://godoc.org/github.com/google/go-github/github#RepositoriesService.GetLatestRelease

exp/examine/local.go
```diff
@@ -0,0 +1,164 @@
...
+		if !(len(queue) > 0) {
+			break
+		}
+	}
+
+	return submodules, nil
+}
+
+//getSubmodules retrieves all submodules related to the root directory
+func getSubmodules(fs afero.Fs, dir string) ([]ModuleWrapper, error) {
```
ryanking: I think we should avoid trying to implement this ourselves, and link to the code in terraform itself to do it.

exp/examine/local.go
```diff
@@ -0,0 +1,164 @@
...
+		if strings.HasPrefix(source.moduleSource, githubURL) { // If the module is a github link, specifically cztack
+			mod, err := GetFromGithub(fs, source.moduleSource)
+			if err != nil {
+				return nil, err
+			}
+
+			sources[i].module = mod
+			sources[i].version = getVersion(sources[i])
+			modules = append(modules, sources[i])
+		} else if strings.HasPrefix(source.moduleSource, awsRegistry) { // If the module is from the tf registry
```
ryanking: I don't think we need to download modules. It's not 100% clear yet but it looks like you're trying to recursively find all modules. If that's the case we can simplify things– since we can't update nested module sources, there is no need to fetch them and try to calculate updates.

exp/examine/latest.go
```diff
@@ -0,0 +1,94 @@
...
+			if err != nil {
+				return nil, errs.WrapUserf(err, "Could not generate url for %s latest module", mod.moduleSource)
+			}
+			module.module, err = GetFromGithub(fs, resource)
+			if err != nil {
+				return nil, err
+			}
+
+			latestModules = append(latestModules, module)
+		} else if strings.HasPrefix(module.moduleSource, awsRegistry) && false { //DISABLED
```
ryanking: Let's remove dead code.

exp/examine/local.go
shedninja6: I guess my thought was that those nested modules might give us more information on what to update.

exp/examine/latest.go
shedninja6: Can't believe I missed this, thank you!

PR #330, opened 2019-07-29
[breaking] Upgrade libraries to TF 0.12

mbarrien:
### Summary
This PR makes Fogg support Terraform 0.12 syntax (and only 0.12 syntax). The different HCL2 syntax makes for a required breaking change to properly support it.

### Test Plan
Passes existing unit tests.


testdata/bless_provider/terraform/accounts/foo/fogg.tf
```diff
@@ -1,12 +1,20 @@
 # Auto-generated by fogg. Do not edit
 # Make improvements in fogg, so that everyone can benefit.
 
+
```
ryanking: why do we get all these blank lines now?

testdata/v1_full/terraform/envs/stage/cloud-env/main.tf
```diff
@@ -5,4 +5,5 @@ module "fogg-test-module" {
   source = "github.com/chanzuckerberg/fogg-test-module"
   bar    = "${local.bar}"
   foo    = "${local.foo}"
+
```
ryanking: is this a change in tf fmt?

testdata/bless_provider/terraform/accounts/foo/fogg.tf
mbarrien: Terraform fmt seems to add way more blank lines. Don't know why, even though the open issue says it won't add new lines.

https://github.com/hashicorp/terraform/issues/21360#issuecomment-493807349

testdata/v1_full/terraform/envs/stage/cloud-env/main.tf
mbarrien: Same as above.

testdata/bless_provider/terraform/accounts/foo/fogg.tf
ryanking: I think that comment is mislead, it seems that they only care about horizontal spacing.

PR #338, opened 2019-08-02
[fix] Fogg init version will generate the latest version by default

shedninja6:
### Summary
Previously `fogg init` defaulted to Fogg version 0, which caused problems with other fogg commands. For example, `fogg plan` would not run because it would read the `fogg.yml` even though it expected a `.json` file which would cause it to error out.
 
### Test Plan
I added tests for the init command so that future features are compatible with `fogg init`.
- Added test to make sure the latest version is always initialized
- Added test to make sure that other versions (specifically 0) can be created


init/init_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	err = writeConfig(fs, conf)
+	a.NoError(err)
+
+	exists, err := afero.Exists(fs, "fogg.yml")
+	a.NoError(err)
+	a.True(exists)
+}
+
+func TestLatestInitVersion(t *testing.T) {
```
ryanking: This test seem to duplicate the one above?

config/config_test.go
```diff
@@ -17,7 +17,7 @@ import (
 
 func TestInitConfig(t *testing.T) {
 	a := assert.New(t)
-	c := InitConfig("proj", "reg", "buck", "table", "prof", "me@foo.example", "0.99.0")
+	c := InitConfig("proj", "reg", "buck", "table", "prof", "me@foo.example", "0.99.0", 0)
```
ryanking: We shouldn't use 0, the first version was version 1 and if there is nothing in the config file (only valid in json) then we assume version 1.

init/init_test.go
shedninja6: Didn't notice the duplication. Thank you!

init/init.go
```diff
@@ -10,6 +10,7 @@ import (
 )
 
 const AWSProviderVersion = "1.27.0"
+const foggVersion = 2
```
edulop91: defaultFoggVersion maybe

init/init.go
```diff
@@ -39,7 +40,7 @@ func writeConfig(fs afero.Fs, config *v2.Config) error {
 //Init reads user console input and generates a fogg.yml file
 func Init(fs afero.Fs) error {
 	project, region, bucket, table, profile, owner := userPrompt()
-	config := config.InitConfig(project, region, bucket, table, profile, owner, AWSProviderVersion)
+	config := config.InitConfig(project, region, bucket, table, profile, owner, AWSProviderVersion, foggVersion)
```
edulop91: I feel like if you're just passing in a constant value to this function it might just be worth it to move that constant over to the config package and reference it directly from there without it having to be a function argument.

config/config.go
```diff
@@ -10,16 +10,16 @@ import (
...
 	"github.com/chanzuckerberg/fogg/errs"
 	"github.com/chanzuckerberg/fogg/util"
+	goVersion "github.com/hashicorp/go-version"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/afero"
 	"gopkg.in/yaml.v3"
-	goVersion "github.com/hashicorp/go-version"
 )
 
 var defaultTerraformVersion = goVersion.Must(goVersion.NewVersion("0.12.5"))
```
edulop91: I would probably move the `defaultFoggVersion` const over here, along with the `defaultTerraformVersion`

shedninja6: Gotcha, thank you!

PR #339, opened 2019-08-02
support github providers

### LINKED ISSUE #239: support github providers

ryanking:
This will enable fogg to configure github providers for you.

### Test Plan
* unit tests

### References
* Fixes #239


config/v2/resolvers.go
```diff
@@ -114,6 +114,22 @@ func ResolveAWSProvider(commons ...Common) *AWSProvider {
 	return nil
 }
 
+// ResolveGithubProvider will return an GithubProvder iff one of the required fields is set somewhere in the set of Common
+// config objects passed in. Otherwise it will return nil.
+func ResolveGithubProvider(commons ...Common) *GithubProvider {
+	org := lastNonNil(GithubProviderOrganizationGetter, commons...)
+
```
edulop91: I think short-ciruiting makes these easier to read
```
if org == nil {
return nil
}
// else actually do stuff
```

templates/common/github_provider.tmpl
```diff
@@ -0,0 +1,10 @@
+{{define "github_provider"}}
+
+provider "github" {
```
edulop91: should we start adding a version requirement on these provider templates? or allowing for one? I've ran into some kind of insidious bugs because I under-specified the bless provider version

ryanking: good point

PR #353, opened 2019-08-21
make setup instead of inline download.sh in .travis.yml

mbarrien:
Fixes issues with git status --porcelain check.


mbarrien:
Changes made... is .fogg/bin in the path?

mbarrien:
Added support to make setup to pass in a fogg version via FOGG_VERSION environment variable. Please re-review.

Alternatively, we could make it read .fogg_version file instead.

PR #365, opened 2019-09-30
implemented github-registry push in goreleaser

kuannie1:


.goreleaser.prerelease.yml
```diff
@@ -21,4 +21,30 @@ archives:
...
 
 release:
-  prerelease: true
\ No newline at end of file
+  prerelease: true
+
+dockers:
+  - dockerfile: Dockerfile
+    image_templates:
+      - docker.pkg.github.com/chanzuckerberg/fogg/fogg:{{.ShortCommit}}
```
edulop91: I wonder if it is worth adding a comment as to why you are using the short commit instead of the version

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+COPY templates templates
+COPY util util
+
+# Build the Go app
+RUN CGO_ENABLED=0 GOOS=linux go build -o fogg .
+
+# Install chamber
+ENV CHAMBER_VERSION=v2.1.0
+RUN wget -q https://github.com//segmentio/chamber/releases/download/${CHAMBER_VERSION}/chamber-${CHAMBER_VERSION}-linux-amd64 -O /bin/chamber
+RUN chmod +x /bin/chamber
```
edulop91: We probably don't need chamber here

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+# Final stage: the running container
+FROM alpine:latest AS final
+
+# Install SSL root certificates
+RUN apk update && apk --no-cache add ca-certificates
+
+COPY --from=builder /app/fogg /bin/fogg
+COPY --from=builder /bin/chamber /bin/chamber
```
edulop91: same here, I would probably get rid of chamber

PR #366, opened 2019-10-08
[feature] simple pre-flight auth checks

ryanking:
Before running any terraform commands, run a quick authentication check.

This won't catch every problem (like authorization failures), but in the cases where a profile is not configured at all, this should give a better error message than terraform does.

Thanks to @kislyuk  for pushing for a better experience here.

Note that his assumes the aws cli is installed. Do we think that is a fair assumption?

### Test Plan
* make test
* running a prerelease on shared-infra

### References


testdata/bless_provider/scripts/component.mk
```diff
@@ -42,11 +42,16 @@ lint-terraform-fmt: terraform
 	@echo
 .PHONY: lint-terraform-fmt
 
+check-auth:
+	@aws --profile $(AWS_BACKEND_PROFILE) sts get-caller-identity >/dev/null || echo "AWS AUTH error. This component is configured to use a profile name $(AWS_BACKEND_PROFILE). Please add one to your ~/.aws/config"
+	@aws --profile $(AWS_PROVIDER_PROFILE) sts get-caller-identity >/dev/null || echo "AWS AUTH error. This component is configured to use a profile name $(AWS_PROVIDER_PROFILE). Please add one to your ~/.aws/config"
```
edulop91: what if we have components without an aws provider configured?

ryanking: This should handle it https://github.com/chanzuckerberg/fogg/pull/366/files#diff-d1dfe7c391918134f1692e607df2c103R9 ?

ryanking: I will test on snowflake-infra

ryanking: derp nevermind, still won't work

ryanking: fixed

templates/repo/scripts/component.mk
```diff
@@ -42,11 +42,17 @@ lint-terraform-fmt: terraform
 	@echo
 .PHONY: lint-terraform-fmt
 
+check-auth:
+	@for p in $(AWS_BACKEND_PROFILE) $(AWS_PROVIDER_PROFILE); do \
```
edulop91: :+1: nice

PR #369, opened 2019-10-15
[feature] fogg exp entropy: an experimental command to instrument entropy in terraform projects

edulop91:
### Summary
- We generate output in logfmt format which is compatible with 
- ~Still missing connecting this with buildevents, but wanted to get some feedback~
- ~One potentially bad thing is we build against a [specific version of TF](https://github.com/chanzuckerberg/fogg/blob/5f26cbf5db0578206b029744f3674896b7b6331a/vendor/github.com/hashicorp/terraform/plans/planfile/tfplan.go#L48), therefore the plan we examine must have been generated using the same tf verison. Some potential friction there unless we're methodically upgrading terraform/fogg versions.~ Seems to be fixed starting in 0.12.10. The only requirement is we upgrade fogg terraform versions promptly since now we can run this command against all plans generated from previous versions of terraform.

### Test Plan
- introduced some changes
- generated a plan
- manually ran the command
``


cmd/exp/entropy.go
```diff
@@ -0,0 +1,100 @@
...
+			err = encoder.EncodeKeyvals(
+				"addr", addr,
+				"resource_mode", resourceMode,
+				"action", action,
+				"component", component,
+				"project", project)
+			if err != nil {
+				return fmt.Errorf("could not encode key/val: %w", err)
+			}
+			err = encoder.EndRecord()
```
ryanking: Have you tested that having one k/v line works? The docs seem to indicate yes.

edulop91: I;ve yet to test this in conjunction with buildevents. Wanted to merge the buildevents pr first

ryanking:
> * One potentially bad thing is we build against a [specific version of TF](https://github.com/chanzuckerberg/fogg/blob/5f26cbf5db0578206b029744f3674896b7b6331a/vendor/github.com/hashicorp/terraform/plans/planfile/tfplan.go#L48), therefore the plan we examine must have been generated using the same tf verison. Some potential friction there unless we're methodically upgrading terraform/fogg versions.

I agree that this is somewhat worrisome. 

An alternative would be to run `terraform show` to get json out put. See https://www.terraform.io/docs/internals/json-format.html

Going that way might suggest using make/shell to do this, rather than running a fogg command.

edulop91:
This changed with tf 0.12.10 to allow us to parse plans for older versions of terraform rather than just the specific version we build with :tada:

templates/repo/scripts/component.mk
```diff
@@ -102,7 +103,7 @@ endif
 .PHONY: init
 
 check-plan: init check-auth ## run a terraform plan and check that it does not fail
-	@$(terraform_command) plan $(TF_ARGS) -detailed-exitcode -lock=false; \
+	@$(terraform_command) plan $(TF_ARGS) -detailed-exitcode -lock=false -out=$(CHECK_PLANFILE_PATH) ; \
```
ryanking: Do you think we should still output the human-readable plan?

ryanking: i think tf show <planfile> would work for that

edulop91: It looks like it still does it

PR #380, opened 2019-11-21
[feature] CircleCI to run tests

edulop91:
Run tests in circleci


testdata/github_provider/scripts/component.mk
```diff
@@ -25,7 +25,7 @@ fmt: terraform ## format code in this component
 lint: lint-terraform-fmt lint-tflint ## run all linters for this component
 .PHONY: lint
 
-lint-tflint: init ## run the tflint linter for this component
+lint-tflint: ## run the tflint linter for this component
```
edulop91: don't need init for fmt

config/config_test.go
```diff
@@ -126,10 +126,20 @@ func TestUpgradeConfigVersion(t *testing.T) {
...
+							Enabled:        boolptr(true),
+							AWSIAMRoleName: util.StrPtr("travis-role"),
+							TestBuckets:    intptr(13),
+							Buildevents:    boolptr(true),
+						},
+					},
+					CircleCI: &v1.CircleCI{
+						CommonCI: v1.CommonCI{
+							Enabled:        boolptr(true),
+							AWSIAMRoleName: util.StrPtr("travis-role"),
```
mbarrien: I know it's just a test file, but "travis-role" inside the circleci test?

config/v1/config.go
```diff
@@ -125,12 +125,26 @@ type Module struct {
...
 type TravisCI struct {
-	Enabled        *bool   `json:"enabled,omitempty" yaml:"enabled,omitempty"`
-	AWSIAMRoleName *string `json:"aws_iam_role_name" yaml:"aws_iam_role_name,omitempty"`
-	TestBuckets    *int    `json:"test_buckets" yaml:"test_buckets,omitempty"`
-	Command        *string `json:"command,omitempty" yaml:"command,omitempty"`
-	Buildevents    *bool   `json:"buildevents,omitempty" yaml:"buildevents,omitempty"`
+	CommonCI `json:",inline" yaml:",inline"`
+}
+
+type CircleCI struct {
```
mbarrien: Should we be adding CircleCI to the v1 config? It's a deprecated config format; it really should only be in v2.

edulop91: i'd rather be consistent with how we define travisCI rather than split these amongst packages

edulop91: Especially since they share several of the struct definitions

mbarrien: Can't just move it to config/v2/config.go with a reference to v1.CommonCI? Personally, I don't like our v2 relying on v1 hack for Travis right now, but I'd rather not make it worse.

edulop91: > Personally, I don't like our v2 relying on v1 hack for Travis
I really don't see why it would be a hack. Would you rather duplicate the definitions?

plan/plan.go
```diff
@@ -178,7 +231,9 @@ func Eval(c *v2.Config) (*Plan, error) {
 
 	p.Modules = p.buildModules(c)
 	p.Atlantis = p.buildAtlantis()
-	p.TravisCI = p.buildTravisCI(c, v)
+	p.TravisCI = p.buildTravisCIConfig(c, v)
+
```
edulop91:
```suggestion
```

config/v1/config.go
edulop91: Made the changes

PR #396, opened 2020-01-28
[feature] aws cred_process redirects stderr to /dev/tty such that users can see mfa prompts

edulop91:
### Summary
botocore will consume both stderr and stdout for their cred_process. We use this hack to make sure that stderr is displayed back to the user. stdout is still consumed by botocore as normal.

### Test Plan
```
aws sts get-caller-identity --profile XXXXXX
INFO[0001] Requesting MFA. Please complete two-factor authentication with your second device             
INFO[0002] Device: XXXXXXX                               
{
    "Account": "XXXXXXXX", 
    "UserId": "XXXXXX", 
    "Arn": "arn:aws:sts::XXXX:assumed-role/XXXX/XXXX"
}
```

### References
https://github.com/boto/botocore/issues/1348


edulop91:
I'll apply @jgadling feedback in a followup pr

PR #407, opened 2020-02-19
fix: CircleCI config proper parallelism/bucketing

edulop91:
We were not properly doing parallelism/workflows. CircleCI config generated could be DRYer, we can make improvements later.


templates/circleci/.circleci/config.yml.tmpl
```diff
@@ -2,30 +2,29 @@
 # Make improvements in fogg, so that everyone can benefit.
 {{- $circleCI := . }}
 version: 2
-
 working_directory: &working_directory ~/fogg
 
 jobs:
-  build:
+{{- range $idx, $testBucket := $circleCI.TestBuckets }}
+  terraform_{{$idx}}:
```
jgadling: Curious why circle's built-in parallelism won't work here?

edulop91: I guess they have many forms of built=in parallelism. Are you referring to the command line globbing/splitting?

jgadling: Essentially, yes. The splitting by timings in particular looks pretty useful.

edulop91: It would work. We can make that improvement later?

PR #412, opened 2020-03-05
[feature] avoid running fresh on every plan

### LINKED ISSUE #411: don't run refresh each time
Basic idea is that in the component.mk generated by fogg we do this–

1.  pull 'refresh' out to its own target and make it a dependency of plan & apply (and maybe others?)
1.  change the plan, apply, etc targets to -refresh=false
1.  In the refresh target, do something like
  1. read a cache file which stores the time of the last refresh (which is .gitignored)
  1. if now - last refresh > some threshold (60 min?)
    1. run refresh

ryanking:
In order to improve development cycles, this diff makes a change that
will only on refresh before a plan once an hour. Also when `terraform
plan` runs a refresh, the updated state does not get persisted whereas
`terraform refresh` does.

Fixes #411 

### Test Plan
* CI
* tested manually on one repo already

### References
* https://www.terraform.io/docs/commands/refresh.html
* https://www.terraform.io/docs/commands/plan.html#refresh-true


mbarrien:
This feature is causing issues in environment where the user/role running terraform plan only has readonly access to the remote state. Examples of this are where a CI-only IAM role only has readonly access to an S3 bucket; the refresh will try to write to the remote state (guaranteed to since the newly checked out repo will not have a .terraform/refreshed_at file within that component), and since it has only read-only permissions, it lacks permissions to write. This results in:

Error: Failed to write state: failed to upload state: AccessDenied: Access Denied
	status code: 403, request id: 4758350D12D315B5, host id:

PR #426, opened 2020-03-23
[feature] remote backend

### LINKED ISSUE #421: fogg support for tfe backend

ryanking:
This PR adds support for the [Terraform  remote backend](https://www.terraform.io/docs/backends/types/remote.html). Until now we only supported the [s3 backend](https://www.terraform.io/docs/backends/types/s3.html) so this required a fair amount of refactoring to make our code be able to accept both.

In this code we –

1. add a Kind field to Backend which can be either s3 or remote, defaulting to s3
2. all the resolution and validation for that new field
3. a significant refactoring of the templates for generating backends. This was a benefit on its own as it helped discover a few hidden bugs.
4. generate a [.terraformignore file](https://www.terraform.io/docs/backends/types/remote.html#excluding-files-from-upload-with-terraformignore) which will help make remote operations more performant by limiting the amount of data we need to upload.

### Test Plan

### References
* Resolves #421


config/v2/validation_test.go
```diff
@@ -76,13 +77,64 @@ func TestValidateOwnersComponent(t *testing.T) {
...
+				},
+				Project:          util.StrPtr("foo"),
+				TerraformVersion: util.StrPtr("1.1.1"),
+			},
+		},
+	}
+
+	if kind == "remote" {
+		var remote Config
+		copier.Copy(&remote, &base)
```
czibuildbot: **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub>
Error return value of `copier.Copy` is not checked (errcheck)

config/v2/config.go
```diff
@@ -152,12 +152,18 @@ type DatadogProvider struct{}
 
 //Backend is used to configure a terraform backend
 type Backend struct {
```
edulop91: random thought: I assume all the fields in this struct are left as is/at the top level to preserve backwards compatibility.
I wonder if we can add a notion of grouping by doing something like
```
Backend struct {
Kind *string
S3Backend //embedded
TFEBackend //embedded
```
that way we keep the same structure in the json but have some notion of semantic grouping in the code. not sure how I feel about this though, maybe a terrible idea.

config/v2/validation.go
```diff
@@ -91,6 +87,41 @@ func ValidateAWSProvider(p *AWSProvider, component string) error {
...
+// ValidateBackend will check the resolved configuration for a backend and return any errors it
+// finds
+func ValidateBackend(backend *Backend, component string) error {
+	if backend == nil {
+		return nil
+	}
+
+	var errs *multierror.Error
+
+	if backend.Kind == nil {
```
edulop91: wonder if this would read nicer if each of the kind statements returned their errors rather than having elifs and the return all the way at the bottom

config/v2/resolvers.go
```diff
@@ -112,6 +114,36 @@ func ResolveAWSProvider(commons ...Common) *AWSProvider {
 	return nil
 }
 
+// ResolveBackend returns the Backend configuration for a given component, after applying all inheritance rules
+func ResolveBackend(commons ...Common) *Backend {
+	kind := lastNonNil(BackendKindGetter, commons...)
+
+	// This feels like a somewhat hacky way to do this, but not sure of a better place yet
+	if kind == nil {
```
edulop91: another random idea is to have a custom marshal on the backend struct that sets the s3 default and then calls the normal unmarshal (which could override the field)

config/v2/config.go
```diff
@@ -152,12 +152,18 @@ type DatadogProvider struct{}
...
-	Type        *string `json:"type,omitempty" yaml:"type,omitempty" validate:"omitempty,oneof=s3"`
+	Kind *string `json:"kind,omitempty" yaml:"kind,omitempty" validate:"omitempty,oneof=s3 remote"`
```
edulop91: I see we have a type alias to string for `BackendKind` in plan, worth using it here too?

config/v2/validation_test.go
```diff
@@ -75,13 +76,66 @@ func TestValidateOwnersComponent(t *testing.T) {
...
-func TestValidateBackend(t *testing.T) {
+func TestValidateBackends(t *testing.T) {
+
+	var cases = []struct {
+		kind     string
+		genValid bool
+		wantErr  bool
+	}{
+		// {"invalid", false, true},
+		// {"s3", true, false},
```
edulop91: not sure if these should stay or go

config/v2/validation.go
```diff
@@ -91,6 +87,41 @@ func ValidateAWSProvider(p *AWSProvider, component string) error {
 	return errs
 }
 
+// ValidateBackend will check the resolved configuration for a backend and return any errors it
+// finds
+func ValidateBackend(backend *Backend, component string) error {
```
edulop91: does this need validation to make sure only one backend (and corresponding fields) is enabled?

plan/plan.go
```diff
@@ -274,12 +288,29 @@ func (p *Plan) buildAccounts(c *v2.Config) map[string]Account {
 
 		accountPlan.AccountName = name
 		accountPlan.ComponentCommon = resolveComponentCommon(defaults.Common, acct.Common)
-		accountPlan.AllAccounts = resolveAccounts(c.Accounts)
+		if accountPlan.ComponentCommon.Backend.Kind == BackendKindS3 {
+			accountPlan.ComponentCommon.Backend.S3.KeyPath = fmt.Sprintf("terraform/%s/accounts/%s.tfstate", accountPlan.ComponentCommon.Project, name)
+		} else if accountPlan.ComponentCommon.Backend.Kind == BackendKindRemote {
+			accountPlan.ComponentCommon.Backend.Remote.Workspace = fmt.Sprintf("accounts-%s", name)
+		}
```
edulop91: should the `else` be an assertion/panic? in case we introduce a bug or later decide to add another one

plan/plan.go
```diff
@@ -309,7 +340,11 @@ func (p *Plan) buildGlobal(conf *v2.Config) Component {
 	global := conf.Global
 
 	componentPlan.ComponentCommon = resolveComponentCommon(defaults.Common, global.Common)
-
+	if componentPlan.ComponentCommon.Backend.Kind == BackendKindS3 {
+		componentPlan.ComponentCommon.Backend.S3.KeyPath = fmt.Sprintf("terraform/%s/global.tfstate", componentPlan.ComponentCommon.Project)
+	} else if componentPlan.ComponentCommon.Backend.Kind == BackendKindRemote {
```
edulop91: same here on the `else` assertion

templates/common/backends.tmpl
```diff
@@ -0,0 +1,34 @@
...
+    }
+  {{ end }}
+  }
+{{ end }}
+
+{{ define "remote_backend" }}
+  {{ if eq .Kind "s3" }}
+    bucket         = "{{ .S3.Bucket }}"
+    {{ if .S3.DynamoTable }}dynamodb_table = "{{ .S3.DynamoTable }}"{{ end }}
+    key            = "{{ .S3.KeyPath }}"
```
edulop91: probably missing something here but backend and remote_backend look exactly the same except for encrypt = true. Do we need the 2 definitions?

config/v2/config.go
ryanking: Interesting idea, but not sure what I think about it.

config/v2/validation.go
ryanking: yeah sgtm

config/v2/resolvers.go
ryanking: The problem with that is if you have inheritance from A->B and set something explicitly in A, but then our marshal code injects something in B, then you resolution will be wrong.

I don't think you can do this at the config / parsing level at all for that reason.

config/v2/config.go
ryanking: 👍

ryanking: I like the idea, but my only problem with that invalid backend types would fail (panic?) at parsing time rather than validation. I think making that check part of validations makes for a better user experience.

config/v2/validation_test.go
ryanking: Good catch, didn't mean to drop those.

config/v2/validation.go
ryanking: Yeah it does. I had been thinking it was impossible to get to that state, but that is only true at the planning stage.

ryanking: Actually this might not work with the way we have things currently build.

It is unlikely that someone would use things this way, but if you have inheritance like s3 -> remote -> s3 you could end up with all the fields set, but need to ignore some of them.

There may be ways for us to ignore this but I can't think of any that would be non-breaking.

plan/plan.go
ryanking: Yeah. I was working under the belief that we are protected by the type checker, but had not thought about the future-proofing angle.

config/v2/resolvers.go
edulop91: ahh, good catch

templates/common/backends.tmpl
ryanking: Yeah in one case we need `workspace = {` and the other `workspaces {`. :(

edulop91: :sob:

PR #440, opened 2020-03-27
[feature] allow setting provider versions for datadog and heroku

ryanking:
This allows pinning versions of the datadog and heroku providers.

### Test Plan
* CI

### References
*


config/v2/config.go
```diff
@@ -146,9 +146,13 @@ type SnowflakeProvider struct {
 	Version *string `json:"version,omitempty" yaml:"version,omitempty"`
 }
 
-type HerokuProvider struct{}
+type HerokuProvider struct {
+	Version *string `yaml:"version,omitempty"`
```
edulop91: I wonder if we should have a BaseProvider or something like it that we embed into these providers

ryanking: 👍

PR #447, opened 2020-04-02
[fix] Fix generation of remote backends

edulop91:
### Summary
We should only configure remote_states for sibling components of type terraform.
Furthermore, remote some dead JSON code

### Test Plan
unittests


templates/account/fogg.tf.tmpl
```diff
@@ -106,7 +106,7 @@ data terraform_remote_state global {
 {{ $outer := . }}
 {{ range $accountName, $backend := .AccountBackends}}
 {{ if ne $accountName $outer.AccountName }}
-data terraform_remote_state {{ $accountName }} {
+data  {{ $accountName }} {
```
edulop91, line 109:
```suggestion
data  terraform_remote_state {{ $accountName }} {
```

PR #450, opened 2020-04-07
un-vendor dependencies

ryanking:
Given the improvements in the go modules implementation (caching,
dependency resolution) I think we can safely stop vendoring. Or at least
that vendoring is not worth the cost anymore. The cost being the level of
noise in pull requests.

### Test Plan
* CI

### References
*


Makefile
```diff
@@ -4,7 +4,6 @@ DIRTY=false
 # TODO add release flag
 GO_PACKAGE=$(shell go list)
 LDFLAGS=-ldflags "-w -s -X $(GO_PACKAGE)/util.GitSha=${SHA} -X $(GO_PACKAGE)/util.Version=${VERSION} -X $(GO_PACKAGE)/util.Dirty=${DIRTY}"
-export GOFLAGS=-mod=vendor
```
edulop91, line 7: Would it be worth adding a test to make sure the go.mod is up-to-date. I kind of worry deps will be added without getting synchronized to go.mod/sum

ryanking: Sounds good.

PR #452, opened 2020-04-07
[feature] automatically prompt for heroku login

palasha:
For users of heroku SSO, the login session times out fairly quickly.
This means that almost every time we use terraform, whether for a plan
or an apply, we see the "login to heroku" error and need to re-run the
command.

This changes the fogg defaults to prompt for login, inline, so that it's
possible to click the SSO link, authenticate, and continue with the
original `make` command uninterrupted.

For non SSO users, this probably will be a no-op, except on their first
run, if they are also not logged into heroku at that point, which seems
really unlikely.

Test Plan:
make
fogg version to confirm latest updates
switch to fogg managed repo
fogg apply
`git grep "heroku auth"` -> observe new command in templates
`make plan` and see that auth is prompted.


ryanking:
@palasha in cases where we are expecting environment variables (CI jobs), I would expect that heroku `heroku auth:whoami` will fail and `heroku auth:login` will hang forever. Do you know if that is the case?

If so I think we should find a way to fail faster for this case.

palasha:
> cases where we are expecting environment variables (CI jobs)

Great question, I had not considered this.

Assuming this would be an unexpected situation, since a CI environment would likely set a .netrc file or HEROKU_API_KEY env var, would a simple 10-15 second timeout be reasonable? 

This kind of makes sense from a user perspective too, if they forget about the login prompt and leave terraform running could it end up holding the lock file for a long time blocking others from doing work?

ryanking:
> would a simple 10-15 second timeout be reasonable?

Yeah I think so.

palasha:
Updated with a `timeout 15`.

PR #453, opened 2020-04-08
[docs] update command for linux install in README

alldoami:


README.md
```diff
@@ -50,8 +50,9 @@ brew untap chanzuckerberg/fogg
...
-1. Go here: https://github.com/chanzuckerberg/fogg/releases, go to the release you want to download, find the tar file that is correct for your machine, and right click the link. Then click "Copy Link Address".
-2. Run `curl -s https://raw.githubusercontent.com/chanzuckerberg/fogg/master/download.sh | bash -s -- -b PATH VERSION` where PATH is where you want to install and VERSION is the version you want to install. To find the path of your current fogg, you can run `which fogg`. Then use the path that is outputted as PATH in the command.
+1. Go here: https://github.com/chanzuckerberg/fogg/releases to find which version of fogg you want.
+2. Run `sudo rm -r PATH; curl -s https://raw.githubusercontent.com/chanzuckerberg/fogg/master/download.sh | sudo bash -s -- -b PATH-FOLDER VERSION` where PATH is where you want to install fogg and VERSION is the specific release version you want to install (format is vx.yy.z). To find the path of your current fogg, you can run `which fogg` (it will look something like /usr/local/bin/fogg). Then use the path that is outputted as PATH in the command. The PATH-FOLDER is the folder in which fogg will be installed (/usr/local/bin).
```
edulop91, line 54: why sudo?

alldoami: on our dev machines, we have to use sudo, probably not necessary for everyone though.

ryanking: I think @alldoami was testing using /usr/local/bin which would require sudo.

edulop91: yep, we moved it to a `~/.local/bin` that was also in PATH instead

README.md
```diff
@@ -50,8 +50,9 @@ brew untap chanzuckerberg/fogg
...
 
 Instructions on downloading the binary:
-1. Go here: https://github.com/chanzuckerberg/fogg/releases, go to the release you want to download, find the tar file that is correct for your machine, and right click the link. Then click "Copy Link Address".
-2. Run `curl -s https://raw.githubusercontent.com/chanzuckerberg/fogg/master/download.sh | bash -s -- -b PATH VERSION` where PATH is where you want to install and VERSION is the version you want to install. To find the path of your current fogg, you can run `which fogg`. Then use the path that is outputted as PATH in the command.
+1. Go here: https://github.com/chanzuckerberg/fogg/releases to find which version of fogg you want.
+2. Run `rm -r PATH; curl -s https://raw.githubusercontent.com/chanzuckerberg/fogg/master/download.sh | bash -s -- -b PATH-FOLDER VERSION` where PATH is where you want to install fogg and VERSION is the specific release version you want to install (format is vx.yy.z). To find the path of your current fogg, you can run `which fogg`. Then use the path that is outputted as PATH in the command. The PATH-FOLDER is the folder in which fogg will be installed.
```
edulop91, line 54: `rm -r PATH;` why is that necessary? will the download script fail if it finds something there? I think naming this `PATH` is somewhat confusing (and I realize that is how it was before) because when I see `PATH` in some shell script I will first think about the PATH environment variable.

alldoami: ya it will fail if something is there

alldoami: @edulop91 can we merge this?

README.md
```diff
@@ -50,8 +50,9 @@ brew untap chanzuckerberg/fogg
...
 
 Instructions on downloading the binary:
-1. Go here: https://github.com/chanzuckerberg/fogg/releases, go to the release you want to download, find the tar file that is correct for your machine, and right click the link. Then click "Copy Link Address".
-2. Run `curl -s https://raw.githubusercontent.com/chanzuckerberg/fogg/master/download.sh | bash -s -- -b PATH VERSION` where PATH is where you want to install and VERSION is the version you want to install. To find the path of your current fogg, you can run `which fogg`. Then use the path that is outputted as PATH in the command.
+1. Go here: https://github.com/chanzuckerberg/fogg/releases to find which version of fogg you want.
+2. Run `rm -r WHICH_FOGG_PATH; curl -s https://raw.githubusercontent.com/chanzuckerberg/fogg/master/download.sh | bash -s -- -b FOGG_PATH VERSION` where PATH is where you want to install fogg and VERSION is the specific release version you want to install (format is vx.yy.z). To find the path of your current fogg, you can run `which fogg`. Then use the path that is outputted as PATH in the command. The PATH-FOLDER is the folder in which fogg will be installed.
```
edulop91, line 54:
```suggestion
2. Run `rm -r WHICH_FOGG_PATH; curl -s https://raw.githubusercontent.com/chanzuckerberg/fogg/master/download.sh | bash -s -- -b FOGG_PATH VERSION` where FOGG_PATH is where you want to install fogg and VERSION is the specific release version you want to install (format is vx.yy.z). To find the path of your current fogg, you can run `which fogg`. Then use the path that is outputted as WHICH_FOGG_PATH in the command. The FOGG_PATH is the folder in which fogg will be installed.
```

PR #456, opened 2020-04-09
[feature] add support for managing tfe providers

ryanking:
### Test Plan
* CI
* manually tested on 1 internal repo

### References
* https://www.terraform.io/docs/providers/tfe/index.html


plan/plan_test.go
```diff
@@ -199,3 +199,24 @@ func Test_buildTerraformIgnore(t *testing.T) {
...
+
+	plan := buildPlan(t, "tfe_provider_yaml")
+
+	on := func(c ComponentCommon) {
+		r.NotNil(c)
+		r.NotNil(c.Providers.Tfe)
+		r.True(c.Providers.Tfe.Enabled)
+	}
+
+	off := func(c ComponentCommon, enabled bool) {
```
edulop91, line 214: enabled unused

templates/account/fogg.tf.tmpl
```diff
@@ -47,6 +47,11 @@
   {{ template "datadog_provider" .Providers.Datadog }}
 {{ end }}
 
+{{ if .Providers.Tfe }}{{ if .Providers.Tfe.Enabled }}
```
edulop91, line 50: wonder if these conditionals should be inside the tfe template so theyre not repeated every fogg.tf template

plan/plan_test.go
ryanking: went with enabled/disabled

templates/account/fogg.tf.tmpl
ryanking: That makes sense, but given that the rest of the them follow this pattern, I would like to leave it as is for the sake of consistency and address it in a follow-up #462

PR #465, opened 2020-04-21
[fix] skip saving planfiles with backend=remote

### LINKED ISSUE #464: turn off generating a saved plan when in remote backend mode
The remote backend does not support locally saving a plan. We use this when running `make check-plan`.

When the remote backend is configured we need to not try to save the plan locally.

ryanking:
The terraform remote backend does not support saving a plan file
locally (I can't find docs for this, but that's what the CLI tells me). This
means our check-plan implementation cannot save it for analysis.

With this PR we turn off locally saving a plan file when backend_kind =
remote.

### Test Plan
* CI
* tested on 2 interal repos - one that is all remote backend and one that is all s3 backend

### References
* Fixes #464

### DIFF
--- a/templates/repo/scripts/component.mk
+++ b/templates/repo/scripts/component.mk
@@ -71,7 +71,7 @@ refresh:
 .PHONY: refresh
 
 refresh-cached:
-	@last_refresh=`cat .terraform/refreshed_at || echo '0'`; \
+	@last_refresh=`cat .terraform/refreshed_at 2>/dev/null || echo '0'`; \
 	current_time=`date +%s`; \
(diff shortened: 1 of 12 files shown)


templates/repo/scripts/component.mk
```diff
@@ -71,7 +71,7 @@ refresh:
 .PHONY: refresh
 
 refresh-cached:
-	@last_refresh=`cat .terraform/refreshed_at || echo '0'`; \
+	@last_refresh=`cat .terraform/refreshed_at 2>/dev/null || echo '0'`; \
```
ryanking, line 74: This is unrelated but was annoying me.

PR #469, opened 2020-04-22
[feature] support configuring aws providers with roles

### LINKED ISSUE #468: fogg assume_role support

ryanking:
Rather than specifying a profile name, you can now specify a role name.
That role name combined with the account id is used to create the arn
for a role to assume.

Previously we only supported using profile names for configuring aws
providers. This allows late-binding of roles wherein you can use the
exact same code but different aws iam principals in different
circumstances. Example - a read-only role for running a plan check in
CI.

For using Terraform Enterprise we want to switch to specifying a role
rather than trying to thread through an aws cli config file.

Fixes #468

### Test Plan
* [x] CI
* [x] manually tested a few internal repos

### References


config/v2/resolvers.go
```diff
@@ -91,24 +79,57 @@ func ResolveStringMap(getter func(Common) map[string]string, commons ...Common)
...
+	var profile, region, role, version *string
+	var accountId *json.Number
+	var additionalRegions []string
+
+	for _, c := range commons {
+		if c.Providers != nil && c.Providers.AWS != nil {
+			p := c.Providers.AWS
+
+			// Profile and Role are mutually exclusive, so if one is set then we set the other to
+			// nil Our previous validations in validateAWSProviderAuth will assure that they are not
```
ryanking, line 95:
```suggestion
			// nil Our validations in validateAWSProviderAuth will assure that they are not
```

templates/account/Makefile.tmpl
```diff
@@ -6,9 +6,9 @@ export TFLINT_ENABLED := {{ if .TfLint.Enabled }}1{{ else }}0{{ end }}
 export TF_PLUGIN_CACHE_DIR := {{ .PathToRepoRoot }}/.terraform.d/plugin-cache
 export TF_BACKEND_KIND := {{ .Backend.Kind }}
 {{ if eq .Backend.Kind "s3" }}export AWS_BACKEND_PROFILE := {{ .Backend.S3.Profile }}{{ end }}
-{{ if .Providers.AWS }}
+{{ if .Providers.AWS }}{{ if .Providers.AWS.Profile }}
```
ryanking, line 9: `text/template`'s  `and` does not short circuit. :)

PR #475, opened 2020-05-18
[feature] support roles for s3 backends

ryanking:
Enable configuring s3 backends with IAM roles.

Similar to our change to allow specifying roles (rather than profiles) for providers, we allow specifying a role as an alternative to a profile. This will enable cases like TFE where it is hard to configure roles.

### Test Plan
* [x] unit tests
* [x] manual test on shared-infra

### References
*


config/v2/resolvers.go
```diff
@@ -137,20 +137,54 @@ func ResolveAWSProvider(commons ...Common) *AWSProvider {
...
+			if b.DynamoTable != nil {
+				dynamoTable = b.DynamoTable
+			}
+
+			if b.Region != nil {
+				region = b.Region
+			}
+
+			// Profile and Role are mutually exclusive, so if one is set then we set the other to
+			// nil Our validations in validateBackend will assure that they are not both set in the
```
edulop91, line 169:
```suggestion
			// nil Our validations in validateBackend will assure that they are not both set or missing in the
```

PR #477, opened 2020-05-27
[fix] set AWS_SDK_LOAD_CONFIG=1

ryanking:
Setting this makes the go sdk behave the way one expects by reading the
aws config.

### Test Plan
* CI

### References
* https://docs.aws.amazon.com/sdk-for-go/api/aws/session/

### DIFF
--- a/templates/repo/scripts/common.mk
+++ b/templates/repo/scripts/common.mk
@@ -6,7 +6,7 @@ TF_VARS := $(patsubst %,-e%,$(filter TF_VAR_%,$(.VARIABLES)))
 REPO_ROOT := $(shell git rev-parse --show-toplevel)
 REPO_RELATIVE_PATH := $(shell git rev-parse --show-prefix)
 AUTO_APPROVE := false
-AWS_SDK_LOAD_CONFIG := 1
+export AWS_SDK_LOAD_CONFIG := 1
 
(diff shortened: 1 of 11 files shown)


jgadling:
This is a behavior change -- is it likely to cause problems for anyone who was depending on the previous behavior? Do we need to provide a config option to disable this change?

ryanking:
> This is a behavior change -- is it likely to cause problems for anyone who was depending on the previous behavior? Do we need to provide a config option to disable this change?

In my experience, no - it seems that everyone who uses the go aws sdk eventually has to set this.

I don't think we need a config option because the `:=` means "set if not set" so to disable you can do `AWS_SDK_LOAD_CONFIG=0 make ..`

templates/repo/scripts/common.mk
```diff
@@ -6,6 +6,7 @@ TF_VARS := $(patsubst %,-e%,$(filter TF_VAR_%,$(.VARIABLES)))
 REPO_ROOT := $(shell git rev-parse --show-toplevel)
 REPO_RELATIVE_PATH := $(shell git rev-parse --show-prefix)
 AUTO_APPROVE := false
+AWS_SDK_LOAD_CONFIG := 1
```
edulop91, line 9:
```suggestion
export AWS_SDK_LOAD_CONFIG := 1
```
I think we have to do that, but I'm not 100% sure

PR #492, opened 2020-08-11
[feature] templates: type var.tags as an object rather than a map

edulop91:
### Summary
Change the automatically generated var.tags from a type of `map(string)` to an `object` type. This lets us make stronger assertions in modules that might want to accept these tag objects.

I found this while trying to write a new module:
```
variable tags {
  description = "Standard tags. Typically generated by fogg."
  # type        = object({ env : string, owner : string, project : string, service : string })
  # NOTE: fogg.tf types this as a map(string) rather than an object. Have to fix there first.
  type = map(string)
}
```

### Test Plan
Unit tests

### References

### DIFF
--- a/templates/account/fogg.tf.tmpl
+++ b/templates/account/fogg.tf.tmpl
@@ -73,7 +73,7 @@ variable account {
 }
 
 variable tags {
-  type =  map(string)
+  type =  object({project: string, env: string, service: string, owner: string, managedBy: string})
   default = {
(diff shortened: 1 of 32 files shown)


ryanking:
@edulop91 can you manually test this on at least 1 repo?

edulop91:
> @edulop91 can you manually test this on at least 1 repo?

verified in an internal repo in the new module i'm working on

PR #495, opened 2020-09-11
[fix] broken terraform.d symlinks

ryanking:
Avoid having broken terraform.d symlinks.

We download custom plugins to the terraform.d directory at the root of the fogg-managed repos and symlink that directory to each working directory. Until recently it was fine if some of these symlinks were broken, but now TFE complains. 

This PR fixes our .gitignore and .terraformignore rules to be sure that there are no broken links.

### Test Plan
* CI
* tested manually via prerelease on 1 repo

### References
*


ryanking:
Going to force-merge this because it is currently breaking many repos.

PR #497, opened 2020-09-25
[feature ] allow limiting remote state configurations

ryanking:
Add `remote_states` to all components (except global) which will filter
the remote state configurations that are generated. Null means to
default to the current behavior.

PR also includes commit intended for the previous release.

### Test Plan
* ci

### References


edulop91 (approved):

config/v2/config.go
```diff
@@ -56,6 +56,7 @@ type Common struct {
 	Owner            *string           `yaml:"owner,omitempty"`
 	Project          *string           `yaml:"project,omitempty"`
 	Providers        *Providers        `yaml:"providers,omitempty"`
+	RemoteStates     []string          `yaml:"remote_states,omitempty"`
```
jgadling, line 59: I think we want the ability to limit account remote state as well as remote states. I was thinking something like:

```
depends_on:
  accounts:
  - foo
  - bar
  components:
  - baz
```

jgadling (changes requested):

config/v2/config.go
ryanking: My implementation will do that (because we use the same list to filter both, though the interface is probably not clear. I like your interface better.

plan/plan.go
```diff
@@ -234,10 +234,11 @@ type Account struct {
 type Component struct {
 	ComponentCommon `yaml:",inline"`
 
-	Accounts  map[string]Account `yaml:"accounts"` // Reference accounts for remote state
-	Component string             `yaml:"component"`
-	EKS       *v2.EKSConfig      `yaml:"eks,omitempty"`
-	Env       string
+	Accounts        map[string]Account `yaml:"accounts"` // Reference accounts for remote state
+	AccountBackends map[string]Backend `yaml:"account_backends"`
```
jgadling, line 238: Maybe it's just me, but this name is a little confusing -- it's RemoteStateBackends rather than AccountBackends?

jgadling (approved):
Would still like to see var naming clarified, but it's not blocking.

ryanking:
@jgadling var naming?

plan/plan.go
ryanking: Yeah it is, but needs a bit of a refactor to fix. I am trying to avoid that refactor for now.

PR #498, opened 2020-09-25
[feature] auto remote state (experimental)

ryanking:
Read a directory of terraform code, find the remote state references and update depends_on in fogg.yml with them.

This code is *very* rough but working for the happy paths. I till do a full parse of code in a terraform directory (in a fogg-managed repo) and extract all the references to `data.terraform_remote_state.foo`. Those will then be used to update depends_on configurations in fogg.yml. It is expected that the user would then run `fogg apply` to update.

Example run, using [dyff](https://github.com/homeport/dyff) to visualize the yaml diff -

```
$ $GOPATH/bin/fogg exp auto-remote-state --path terraform/accounts/czi-id
$ git diff

accounts.czi-id
  + one map entry added:
    depends_on:
    │ accounts:
    │ - okta-czi-duo
```

### Test Plan

I've tested this manually on a few repos and components, working to do some more.

### References


cmd/root.go
```diff
@@ -38,6 +36,24 @@ var rootCmd = &cobra.Command{
 			}
 			return pprof.StartCPUProfile(f)
 		}
+
+		debug, err := cmd.Flags().GetBool("debug")
```
ryanking, line 40: refactored here to make it easier to use across all commands

config/v2/config.go
```diff
@@ -39,6 +39,20 @@ func ReadConfig(fs afero.Fs, b []byte, configFile string) (*Config, error) {
 	return c, e
 }
 
+func (c *Config) Write(fs afero.Fs, path string) error {
```
ryanking, line 42: moved from init module

config/v2/config.go
```diff
@@ -79,7 +93,7 @@ type Tools struct {
 type CircleCI struct {
 	CommonCI `yaml:",inline"`
 
-	SSHKeyFingerprints []string `yaml:"ssh_key_fingerprints"`
+	SSHKeyFingerprints []string `yaml:"ssh_key_fingerprints,omitempty"`
```
ryanking, line 96: was showing up via Config.Write

exp/state/remote_state.go
```diff
@@ -0,0 +1,223 @@
...
+				components = append(components, r)
+			}
+		}
+	}
+
+	// update fogg.yml with new references
+	logrus.Debugf("found accounts %#v", accounts)
+	logrus.Debugf("found components %#v", components)
+
+	if componentType == "accounts" {
```
edulop91, line 65: maybe make this a switch and error on the default, just in case

exp/state/remote_state.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+	}
+
+	return conf.Write(fs, configFile)
+}
+
+func collectRemoteStateReferences(path string) ([]string, error) {
+	fs := tfconfig.NewOsFs()
+
+	references := map[string]bool{}
```
edulop91, line 96: https://github.com/chanzuckerberg/go-misc/blob/main/sets/strings.go, either here or in the future

exp/state/remote_state.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+
+		name := info.Name()
+		ext := fileExt(name)
+		if ext == "" || isIgnoredFile(name) {
+			continue
+		}
+
+		baseName := name[:len(name)-len(ext)] // strip extension
+		isOverride := baseName == "override" || strings.HasSuffix(baseName, "_override")
```
edulop91, line 188: huh, I didn't know about this behavior in TF

exp/state/remote_state.go
```diff
@@ -0,0 +1,223 @@
...
+		for _, block := range content.Blocks {
+			logrus.Debugf("block type: %v", block.Type)
+
+			attrs, _ := block.Body.JustAttributes()
+
+			for _, v := range attrs {
+				refs, _ := lang.ReferencesInExpr(v.Expr)
+
+				for _, r := range refs {
+					if r != nil {
```
edulop91, line 143: I'd invert this to do if `r == nil: continue` so you have flatter loop

edulop91 (approved):
this is great. left some comments that I think might make this easier to read/understand.

exp/state/remote_state.go
ryanking: yeah it is weird

config/v2/config.go
```diff
@@ -244,6 +260,38 @@ func (ck *ComponentKind) GetOrDefault() ComponentKind {
 	return *ck
 }
 
+// PathToComponentType given a path, return information about that component
+func (c Config) PathToComponentType(path string) (string, string, string, error) {
```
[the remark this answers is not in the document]
ryanking: Good idea. I was bit annoyed by this but didn't have a good idea for a better way.

PR #506, opened 2020-11-19
[exp] module-map command

ryanking:
Add an experimental command that, given a path will list all the local module dependencies. Will be useful for doing automation work with TFE.

### Test Plan
* CI
* manually tested on shared-infra


jgadling (commented):
Bumping this to follow local module dependencies and THEIR dependencies down to build up more complete lists would be useful. Even if we have to set a hard limit of how many levels to traverse.

exp/modules/modules.go
```diff
@@ -0,0 +1,101 @@
...
+			attrs, _ := block.Body.JustAttributes()
+			for _, a := range attrs {
+				if a.Name == "source" {
+					v, err := a.Expr.Value(nil)
+					if err != nil {
+						return err
+					}
+
+					// we only want local modules, so use go-getter to detect "file:" type
+					str, err2 := getter.Detect(v.AsString(), path, getter.Detectors)
```
mbarrien, line 73: This is blowing up with references to terraform-aws-modules, which uses a source string that looks like a path, along with a version.

A better way may be this from the Terraform documentation: https://www.terraform.io/docs/language/modules/sources.html#local-paths "A local path must begin with either ./ or ../ to indicate that a local path is intended, to distinguish from a module registry address."

ryanking: @mbarrien can I see an example of what you are talking about? That said searchin for a prefix of `./` or `..` would certainly be easier.

mbarrien: It was run on Jess's machine, so I only saw a screenshot of it; need to reproduce on my machine to get you the exact text. But it was crashing processing a module that referred to "terraform-aws-modules/security-group-aws" and trying to find files in a non-existent directory

ryanking: Ah I think I know what's going on. Registry modules are not handled by go-getter (terraform special-cases them further up the stack). So my go-getter approach was not going to work.

I switch to a simpler model, if you want to take a look.

jgadling (approved):

PR #513, opened 2021-01-13
[feature] command to list component paths

ryanking:
Add a command to list all configured component paths.

### Test Plan
* unit tests

### References
*


cmd/components_paths.go
```diff
@@ -0,0 +1,64 @@
...
+	componentsCmd.AddCommand(componentsListCmd)
+	componentsListCmd.Flags().StringP("config", "c", "fogg.yml", "Use this to override the fogg config file.")
+}
+
+var componentsListCmd = &cobra.Command{
+	Use:           "paths",
+	Short:         "List paths for all configured components.",
+	SilenceErrors: true, // If we don't silence here, cobra will print them. But we want to do that in cmd/root.go
+	RunE: func(cmd *cobra.Command, args []string) error {
+		var e error
```
edulop91, line 24:
```suggestion
```
should this for consistency with the other var definitions?

cmd/components_paths.go
```diff
@@ -0,0 +1,64 @@
...
+		// Set up fs
+		pwd, e := os.Getwd()
```
edulop91, line 26:
```suggestion
		pwd, err := os.Getwd()
```
I think `err` is more "idiomatic" + consistent with our go code

cmd/components_paths.go
```diff
@@ -0,0 +1,64 @@
...
+
+		for _, a := range p.Accounts {
+			fmt.Printf("terraform/%s\n", a.AccountName)
+		}
+
+		for _, e := range p.Envs {
+			for _, c := range e.Components {
+				fmt.Printf("terraform/%s/%s\n", e.Env, c.Component)
+			}
+		}
```
edulop91, line 61: I wonder if this should be a method on the plan?

edulop91 (approved):

cmd/components_paths.go
ryanking: Yeah I was thinking about that, but given that so far we only need it in one place, I decided to YAGNI for now.

edulop91 (approved):

PR #516, opened 2021-01-20
[fix] exp auto-remote-state to handle empty lists correctly

ryanking:
Do not ignore empty lists in the `exp auto-remote-state` command.
Without this change an empty list would not get written to fogg.yml,
meaning that components with zero remote state dependencies would not
get updated.

### Test Plan
* ci
* tested manually on 1 internal repo

### References

### DIFF
--- a/config/v2/config.go
+++ b/config/v2/config.go
@@ -235,8 +235,8 @@ type CommonCI struct {
 }
 
 type DependsOn struct {
-	Accounts   []string `yaml:"accounts,omitempty"`
-	Components []string `yaml:"components,omitempty"`
+	Accounts   []string `yaml:"accounts"`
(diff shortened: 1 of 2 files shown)


config/v2/config.go
```diff
@@ -235,8 +235,8 @@ type CommonCI struct {
 }
 
 type DependsOn struct {
-	Accounts   []string `yaml:"accounts,omitempty"`
-	Components []string `yaml:"components,omitempty"`
+	Accounts   []string `yaml:"accounts"`
+	Components []string `yaml:"components"`
```
ryanking, line 239: needed to generate empty lists from the auto-remote-state command

jgadling (approved):

PR #518, opened 2021-01-22
[feature] fogg fmt - standard formatting for fogg.yml

ryanking:
Add `fogg fmt` command which implements a standard formatting for the fogg.yml file (shared with other commands which update fogg.yml.

Various scripts, code editors and tools format yaml files differently, leading to diffs with a lot of noise. This command will help enforce a standardized format.

### Test Plan
* manually tested on all internal repos

### References
*

### DIFF
--- a/cmd/fmt.go
+++ b/cmd/fmt.go
@@ -1,0 +1,61 @@
+package cmd
+
+import (
+	"os"
+
+	"github.com/chanzuckerberg/fogg/config"
+	"github.com/chanzuckerberg/fogg/errs"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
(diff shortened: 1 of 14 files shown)


mbarrien:
Won't this strip out comments?

ryanking:
> Won't this strip out comments?

Hmm yeah, which means we have been stripping comments out for awhile (via the other commands that update fogg.yml).

ryanking:
@mbarrien after digging on this it seems that we can't have aggressive formatting (re-ordering of keys) and preservation of comments, so I am simplifying this to deal with just indentation. we can get more aggressive later

mbarrien (approved):

PR #522, opened 2021-02-05
[refactor] use same templates for components, accounts and global

ryanking:
Rather than separate and not quite equivalent implementations, use one.

One side effect is that for components,  terraform version constraints are now tighter (= rather than ~>). This was the way accounts were set up and probably the way we should work.

### Test Plan
* CI


apply/apply.go
```diff
@@ -79,7 +80,8 @@ func Apply(fs afero.Fs, conf *v2.Config, tmp *templates.T, upgrade bool) error {
 		return errs.WrapUser(e, "unable to apply envs")
 	}
 
-	e = applyGlobal(fs, p.Global, &tmp.Global, &tmp.Common)
+	x := tmp.Components[v2.ComponentKindTerraform]
+	e = applyGlobal(fs, p.Global, &x, &tmp.Common)
```
edulop91, line 84:
```suggestion
	tfBox = tmp.Components[v2.ComponentKindTerraform]
	e = applyGlobal(fs, p.Global, &tfBox, &tmp.Common)
```
not sure if that makes sense

plan/plan.go
```diff
@@ -224,27 +230,19 @@ type Module struct {
 type Account struct {
 	ComponentCommon `yaml:",inline"`
 
-	AccountBackends map[string]Backend      `yaml:"account_backends"` // Reference accounts for remote state
-	AllAccounts     map[string]*json.Number `yaml:"all_accounts"`
-	AccountName     string                  `yaml:"account_name"`
-	Global          *Component
+	Account string
+	Global  *Component
```
edulop91, line 234: do these need yaml tags? (since the ones removed had them)

plan/plan.go
```diff
@@ -224,27 +230,19 @@ type Module struct {
...
-	Component       string             `yaml:"component"`
-	EKS             *v2.EKSConfig      `yaml:"eks,omitempty"`
-	Env             string
-
-	Kind              *v2.ComponentKind  `yaml:"kind,omitempty"`
-	ModuleSource      *string            `yaml:"module_source"`
-	ModuleName        *string            `yaml:"module_name"`
-	ComponentBackends map[string]Backend `yaml:"component_backends"`
-	Global            *Component         `yaml:"global"`
+	EKS          *v2.EKSConfig     `yaml:"eks,omitempty"`
```
edulop91, line 241: sidenote: I wonder if we can drop this since the AWS provider now has the ability to ignore tags

templates/component/terraform/fogg.tf.tmpl
```diff
@@ -40,7 +40,7 @@
 
 
 terraform {
-  required_version = "~>{{ .TerraformVersion }}"
+  required_version = "={{ .TerraformVersion }}"
```
edulop91, line 43: is that a breaking change? does it matter?

edulop91 (approved):
some small comments

jgadling (changes requested):
There's a reverse-incompatibility from previous fogg versions -- is this intentional?
fogg.yml looks like this:

```
accounts:
  acctname:
    <acct_config>
    depends_on:
      accounts: []
      components: []
```

And the error:

```
ERROR: unable to apply accounts: unable to apply templates to account: unable to apply template: template: terraform/accounts/acctname/fogg.tf:99:16: executing "terraform/accounts/acctname/fogg.tf" at <.Component>: can't evaluate field Component in type plan.Account
```

plan/plan.go
ryanking: Yeah problem, will make plan output easier to read.

templates/component/terraform/fogg.tf.tmpl
ryanking: Breaking in theory, but not in practice if you use the makefiles.

ryanking:
@jgadling not intended, and not able to reproduce

jgadling (approved):

PR #539, opened 2021-02-22
auto-tidy go mods

ryanking:
Add a new GHA workflow that will run `make deps` and push the result
back. Already tested in go-misc.


.github/workflows/mods.yml
```diff
@@ -0,0 +1,21 @@
...
+          token: ${{ secrets.CZIBUILDBOT_GITHUB_TOKEN }}
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.15.2'
+      - name: fix go modules
+        run: |
+          go mod tidy
+          git add go.*
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
edulop91, line 18: not sure about this but, would this work in the "general" case?

ryanking: What do you mean by the general case? Do you mean when there is nothing to do?

edulop91 (approved):

.github/workflows/mods.yml
edulop91: nvm, I got confused

PR #562, opened 2021-03-31
[fix] Remove git checkout HEAD^2  in CodeQL initialization process

kuannie1:
Fix based on a warning from [this build](https://github.com/chanzuckerberg/go-misc/runs/2238808342?check_suite_focus=true):
```
Warning: 1 issue was detected with this workflow: git checkout HEAD^2 is no longer necessary. Please remove this step as Code Scanning recommends analyzing the merge commit for best results.
```


kuannie1:
merging because this change will help us with debugging other failing builds.

PR #566, opened 2021-04-02
Add support for ARM architectures

damien:
### Summary

1. Adds ARM64 as a compile target (Apple Silicon, possibly RPi if you're feeling adventurous).
2. Tweak binary attributes generated by the build to get us a bit closer to a reproducible build.
3. Upgrade packr from [v1](https://github.com/gobuffalo/packr/tree/master) to [v2](https://github.com/gobuffalo/packr/tree/master/v2). Older versions of packr were causing the build to fail on Apple M1, newer versions work without issue. Worth noting that we may want to later follow up by replacing packr with [pkgr](https://github.com/markbates/pkger), as the author of packr recommends doing so and it's unclear how much longer packr will continue to be supported (see https://github.com/gobuffalo/packr/issues/252).

### Test Plan

- [x] unit tests
- [x] local build
- [x] ci build

### Notes

1. This PR addresses _building_ fogg for ARM64. It's still not possible to _release_ from ARM64, as there the release tooling used by fogg doesn't yet support ARM64


.goreleaser.prerelease.yml
```diff
@@ -12,15 +12,17 @@ builds:
       - linux
     goarch:
       - amd64
+      - arm64
     ldflags:
-      - '-w -s -X github.com/chanzuckerberg/fogg/util.GitSha={{.Commit}} -X github.com/chanzuckerberg/fogg/util.Version={{.Version}} -X github.com/chanzuckerberg/fogg/util.Dirty=false -X github.com/chanzuckerberg/fogg/util.Release=true'
+      - '-w -s -X main.Date={{.CommitDate}} -X github.com/chanzuckerberg/fogg/util.GitSha={{.Commit}} -X github.com/chanzuckerberg/fogg/util.Version={{.Version}} -X github.com/chanzuckerberg/fogg/util.Dirty=false -X github.com/chanzuckerberg/fogg/util.Release=true'
```
edulop91, line 17: ` -X main.Date={{.CommitDate}}`

Why this change?

edulop91 (commented):
thank you, looks good! Just one question.

wrt releasing:
> It's still not possible to release from ARM64, as there the release tooling used by fogg doesn't yet support ARM64

I thought I saw a handful of Terraform providers, also using goreleaser, start releasing for for goos=darwin,goarch=arm64. The [AWS](https://github.com/hashicorp/terraform-provider-aws/pull/17703) one for example. Would it be worth following up with a quick PR to enable releases for darwin/arm64?

.goreleaser.prerelease.yml
damien: It was a recommended change on the Goreleaser docs if you want builds to be reproducible from a given commit. Relying on the commit date/time rather than the system time means that we get the _exact_ same metadata on the binary when we build it locally vs on CI.

damien: This may not get us a perfectly reproducible build, but it's one less thing that'll be different from builds of the same commit.

hanxlin (approved):

PR #573, opened 2021-04-16
feat[auth0]: adding auth0 provider to fogg

jacobrheath:
### Summary
This PR adds an Auth0 provider to the fogg configuration so that developers can configure Auth0 terraform using fogg.

### Test Plan
Created new golden files in : https://github.com/chanzuckerberg/fogg/blob/692206a991fc8a5c52334e3a66ace53deb04af33/apply/golden_file_test.go

### References
https://github.com/alexkappa/terraform-provider-auth0


plan/plan.go
```diff
@@ -199,6 +200,11 @@ type GithubProvider struct {
 	BaseURL      *string `yaml:"base_url"`
 }
 
+//TODO: ?
```
edulop91, line 203: rm

edulop91 (approved):

PR #585, opened 2021-07-06
provide Github token to the apply step during CI

alexsmaliy-czi:
### Summary
When we use Fogg CI in a private repo, we are unable to use pinned URL references to the repo itself as values of `module_source`, because Fogg (via go-getter) is not authorized to download them. The minimal fix for this is to populate a GITHUB_TOKEN env var during the `fogg apply` step of Github CI.

### Test Plan
unittests


alexsmaliy-czi:
Closing, since it looks like the approved mechanism is to inject a private key using `defaults.tools.github_actions_ci.ssh_key_secrets` in the config file.

PR #629, opened 2022-04-11
Fix typo

adnanislamczi:
### Summary
A description of changes or issues addressed by this PR. Provide links to relevant PRs if any.

We were having issues uninstalling and reinstalling the most up to date version of fogg and we found that `brew untap chanzuckerberg/tap` is the proper command to use

### Test Plan

Make sure the latest version of fogg gets installed using the updated commands

### References

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ this, then install as above–
 
 ```shell
 brew uninstall fogg
-brew untap chanzuckerberg/fogg
+brew untap chanzuckerberg/tap
 ```
 
 ## Linux


jakeyheath:
I am not a brew expert. What was the difference here? I would imagine that the new command would uninstall all chanzuckerberg taps?

jakeyheath:
I am not a brew expert. What was the difference here? I would imagine that the new command would uninstall all chanzuckerberg taps?

edulop91 (approved):

PR #663, opened 2022-06-13
feat: allow fogg to authenticate with GH OAuth token

jakeyheath:
### Summary
This PR introduces a feature to fogg that allows the user to provide a Github OAuth token when doing module source downloading instead of SSH keys. This is handy for bot users and OAuth automation that use Fogg in the backend and don't have SSH keys. This should allow someone to do the following:

~~~
FOGG_GITHUBTOKEN=<token> fogg apply
~~~

Where the `FOGG_GITHUBTOKEN` is a personal access token or a token received through an OAuth flow.

### Test Plan
* Unit tests

### References


apply/apply.go
```diff
@@ -204,7 +239,19 @@ func applyEnvs(
...
+				downloader := util.MakeDownloader(*componentPlan.ModuleSource)
+				if oauth {
+					type HTTPAuth struct {
+						GithubToken string `required:"true"`
+					}
+					var httpAuth HTTPAuth
+					err := envconfig.Process("fogg", &httpAuth)
+					if err != nil {
+						return errs.WrapUser(err, "unable to get env FOGG_GITHUBTOKEN with the --oauth flag")
+					}
```
edulop91, line 251: should you check this much earlier (like at the cobra validation stage) so that you fail early if the env is not set when the flag is?

cmd/apply.go
```diff
@@ -36,6 +37,11 @@ var applyCmd = &cobra.Command{
 			return e
 		}
 
+		oauth, e := cmd.Flags().GetBool("oauth")
```
edulop91, line 40: --oauth seems a bit ambiguous here. What do you think about not specifying a flag at all and instead having this behavior totally driven by the presence/absence of the env variable?

apply/apply.go
edulop91, line 251: should all of this logic instead be embedded as part of the downloader itself? otherwise users of this code need to be making these decisions each time and it will be out of sync eventually

apply/apply.go
```diff
@@ -165,12 +166,46 @@ func applyModules(fs afero.Fs, p map[string]plan.Module, moduleBox, commonBox fs
...
+		return sURL
+	}
+
+	splits := strings.Split(s, "git::ssh://git@")
+	if len(splits) != 2 {
+		return sURL
+	}
+	u, err := url.Parse(fmt.Sprintf("https://%s", splits[1]))
+	if err != nil {
+		logrus.Error()
```
edulop91, line 193: log the error

apply/apply.go
```diff
@@ -165,12 +166,46 @@ func applyModules(fs afero.Fs, p map[string]plan.Module, moduleBox, commonBox fs
...
+	// only detect the remote destinations
+	s, err := getter.Detect(sURL, token, []getter.Detector{
+		&getter.GitLabDetector{},
+		&getter.GitHubDetector{},
+		&getter.GitDetector{},
+		&getter.BitBucketDetector{},
+		&getter.S3Detector{},
+		&getter.GCSDetector{},
+	})
+	if err != nil {
```
edulop91, line 183: log the error, at least debug

alldoami (approved):

PR #693, opened 2022-08-01
feat: CCIE-180 allow running 'fogg init' with flags instead of realtime user prompts

hspitzley-czi:
### Summary
A description of changes or issues addressed by this PR. Provide links to relevant PRs if any.

### Test Plan

### References

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -11,6 +11,8 @@ import (
 	"github.com/spf13/pflag"
 )
 
+type FoggProject = fogg_init.FoggProject
+
 func init() {
 	initCmd.Flags().String("project", "", "Use this to pass the project name via CLI.")
 	initCmd.Flags().String("region", "", "Use this to pass the aws region via CLI.")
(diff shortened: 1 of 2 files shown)


cmd/init.go
```diff
@@ -5,14 +5,88 @@ import (
...
+			found = true
+		}
+	})
+	return found
+}
+
+func userPrompt(cmd *cobra.Command) (string, string, string, string, string, string, error) {
+	project, region, bucket, table, profile, owner := "", "", "", "", "", ""
+
+	project, e := cmd.Flags().GetString("project")
```
jakeyheath, line 37: I think conventionally use `err` for error variables.

cmd/init.go
```diff
@@ -5,14 +5,88 @@ import (
...
+	found := false
+	cmd.Flags().Visit(func(f *pflag.Flag) {
+		if f.Name == name {
+			found = true
+		}
+	})
+	return found
+}
+
+func userPrompt(cmd *cobra.Command) (string, string, string, string, string, string, error) {
```
jakeyheath, line 34: after 3 output vars, I usually revert to making a struct because I forget the ordering of the variables, especially when they are all of the same type. It might make it a little cleaner to do something like this:

```suggestion
type FoggProject struct {
Project, Region, Bucket, Table, Profile, Owner string
}
func userPrompt(cmd *cobra.Command) (*FoggProject, error) {
```

and then in error cases you'd do something like 

~~~go
if err != nil {
  return nil, errors.Wrap(err, "a summary of the error")
}
~~~

jakeyheath (approved):
minor go conventions

PR #731, opened 2022-09-19
feat: Default module variables

### LINKED ISSUE #163: Default Variables for Modules
fogg generates a blob in main.tf in the respective env after I `fogg apply`. If I specify the `module_source` for that component, it is going to generate a parameter for each variable defined in the module. For example 

```
module "my-default-module " {
  source                 = "../../../modules/my-default-module"
  env                    = "${local.env}"
}
```

I have to define `local.env` in another file within this component. 

However, fogg still generates a parameter for me if one of the variable has a default value. I will have to either delete that line in main.tf (which is auto-generated) or redeclare that variable with the default value (It kind of defy the purpose of setting up the default).

vincenthsh:
- fix: Upgrade terraform-config-inspect
- feat: Add optional variables to component conf

### Summary
forked off my branch with `terraform-config-inspect` upgrade.

Allows users to supply a list of variables for a module at the component level, to fix #163 

### Test Plan
Unit test: `TestApplyModuleInvocationWithEmptyVariables`
Unit test: `TestApplyModuleInvocationWithOneDefaultVariable`

### Example

```yaml
envs:
  shared-services:
    components:
      state-s3-backend:
        module_name: backend
        module_source: "github.com/nozaq/terraform-aws-remote-state-s3-backend?ref=v1.3.3"
        variables:
          - override_s3_bucket_name
```

results in `main.tf`:

```hcl
# Auto-generated by fogg. Do not edit

module "backend" {
  source                  = "github.com/nozaq/terraform-aws-remote-state-s3-backend?ref=v1.3.3"
  override_s3_bucket_name = local.override_s3_bucket_name

}
```

passing in `variables:[]` will only generate placeholders for module required variables.

omitting `variables` all together results in HEAD behavior:

```hcl
# Auto-generated by fogg. Do not edit

module "backend" {
  source                                 = "github.com/nozaq/terraform-aws-remote-state-s3-backend?ref=v1.3.3"
  dynamodb_enable_server_side_encryption = local.dynamodb_enable_server_side_encryption
  dynamodb_table_billing_mode            = local.dynamodb_table_billing_mode
  dynamodb_table_name                    = local.dynamodb_table_name
  enable_replication                     = local.enable_replication
  iam_policy_attachment_name             = local.iam_policy_attachment_name
  iam_policy_name_prefix                 = local.iam_policy_name_prefix
  iam_role_arn                           = local.iam_role_arn
  iam_role_name_prefix                   = local.iam_role_name_prefix
  iam_role_permissions_boundary          = local.iam_role_permissions_boundary
  kms_key_alias                          = local.kms_key_alias
  kms_key_deletion_window_in_days        = local.kms_key_deletion_window_in_days
  kms_key_description                    = local.kms_key_description
  kms_key_enable_key_rotation            = local.kms_key_enable_key_rotation
  noncurrent_version_expiration          = local.noncurrent_version_expiration
  noncurrent_version_transitions         = local.noncurrent_version_transitions
  override_s3_bucket_name                = local.override_s3_bucket_name
  replica_bucket_prefix                  = local.replica_bucket_prefix
  s3_bucket_force_destroy                = local.s3_bucket_force_destroy
  s3_bucket_name                         = local.s3_bucket_name
  s3_bucket_name_replica                 = local.s3_bucket_name_replica
  s3_logging_target_bucket               = local.s3_logging_target_bucket
  s3_logging_target_prefix               = local.s3_logging_target_prefix
  state_bucket_prefix                    = local.state_bucket_prefix
  tags                                   = local.tags
  terraform_iam_policy_create            = local.terraform_iam_policy_create
  terraform_iam_policy_name_prefix       = local.terraform_iam_policy_name_prefix

}
```


jakeyheath:
Thanks for working on this feature! I guess this would fix it, although the user would need to know all the default variables and place them in their yaml file, which seems only marginally better. Ideally, we want to take advantage of the defaults that the terraform module provides so that we don't need to add some much stuff to the fogg.yml. 

Could we update this so that if a variables key is provided it does:

- for an `[]` variables, it only autogenerates the required variable arguments
- for an `[stringa, stringb, ...]` variables, it autogenerates the required variables arguments, with each string as an override of a default
- for no variables key, it defaults to the HEAD behavior to preserve backward compatability

jakeyheath:
Thanks for working on this feature! I guess this would fix it, although the user would need to know all the default variables and place them in their yaml file, which seems only marginally better. Ideally, we want to take advantage of the defaults that the terraform module provides so that we don't need to add some much stuff to the fogg.yml. 

Could we update this so that if a variables key is provided it does:

- for an `[]` variables, it only autogenerates the required variable arguments
- for an `[stringa, stringb, ...]` variables, it autogenerates the required variables arguments, with each string as an override of a default
- for no variables key, it defaults to the HEAD behavior to preserve backward compatability

vincenthsh:
>  although the user would need to know all the default variables and place them in their yaml file, which seems only marginally better.

the opposite actually, the user would need to choose which variables he wants specifically (either required or overwriting defaults)

from my experience, most modules have a limited set of required variables.. an option could be to make the code include any variables that are required as well (on top of what the user lists.. ) but it would a bit disjointed

vincenthsh:
>Could we update this so that if a variables key is provided it does:
>
> - for an [] variables, it only autogenerates the required variable arguments
> - for an [stringa, stringb, ...] variables, it autogenerates the required variables arguments, with each string as any variable on the module
> - for no variables key, it defaults to the HEAD behavior to preserve backward compatability

right, adding the required, I also just realized, would be quite a good thing to add..

I'll have to see how I'd diff variables omitted (nil array) vs variables provided with empty list ...

I think easiest would just change the code to still loop through the vars and see if any required variable isn't in the list the user provided (I find that in this case, the users generated code is a bit mystical... I'll work on this

vincenthsh:
been playing with this, and it's quite annoying that it's only possible to limit to required vars by at least specifying one required var (and it will pick up all the others) ... I will see if the config parsing logic in plan.go can diffirentiate between not passing in the `variables` and passing in an empty list, any hints?

vincenthsh:
I'll need to spend more time on understanding https://github.com/go-yaml/yaml/commit/bef53efd0c76e49e6de55ead051f886bea7e9420 to evaluate if the `variables` keyword was omitted or set to an emtpy list

vincenthsh:
Now it seems to handle all scenarios udpated test cases:

test-module has:
- required var foo
- required var bar
- optional var baz
- optional var quux

Test Cases (these should really be refactored into [table driven tests](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) - but my initial PR tried to make minimal changes ..

- `TestApplyModuleInvocation` and `TestApplyModuleInvocationWithModuleName` test HEAD behavior (`variables` is `nil`)
- `TestApplyModuleInvocationWithEmptyVariables` tests `variables: []` - only placeholders for foo and bar are expected
- `TestApplyModuleInvocationWithOneDefaultVariable` tests `variables: ['baz']` - placeholders for `foo`, `bar` and `baz` are expected (sorted)

jakeyheath (approved):
sick!

PR #846, opened 2023-05-05
fix: use Terraform-supported tfenv instead of ours

kuannie1:
### Summary
Before we've been using our own version of this tool, but I've been getting [this error](https://support.hashicorp.com/hc/en-us/articles/13177506317203#:~:text=If%20you%20run%20a%20previously%20downloaded%20Apple%20artifact%20signed%20with%20the%20old%20certificate%2C%20you%20may%20see%20an%20error%20like%3A). They have recommendations, but the easiest thing I could think of is upgrading this tfenv dependency to the latest supported version. By manually tweaking this value, I could run `make apply` and `make plan` without any blocking popups. 

The only difference between the czi version and the official one is [this change](https://github.com/chanzuckerberg/tfenv/pull/1), but I don't know if this change matters or not. If it doesn't and we use the tfutils repo, we can archive the czi repo.

### Test Plan

### References
https://support.hashicorp.com/hc/en-us/articles/13177506317203
https://www.storagetutorials.com/solved-terraform-will-damage-your-computer/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+### [0.83.1](https://github.com/chanzuckerberg/fogg/compare/v0.83.0...v0.83.1) (2023-05-08)
+
+
+### Misc
+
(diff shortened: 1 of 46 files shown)


jakeyheath (approved):

jakeyheath:
Wow we have so many forks! Good catch

PR #873, opened 2023-06-02
feat: Fogg template for k8s

alexlokshin-czi:
To generate a k8s template, configure `kubernetes` provider like this (you need to specify the `cluster_name`):

```yaml
providers:
  kubernetes:
    enabled: true
    cluster_name: eks-v2
    version: 2.19.0
```


templates/templates/common/kubernetes_provider.tmpl
```diff
@@ -0,0 +1,15 @@
+{{ define "kubernetes_provider" }}
+{{ if .ClusterName }}
```
jakeyheath, line 2:
```suggestion
{{ define "kubernetes_provider" -}}
{{ if .ClusterName -}}
```

jakeyheath (approved):
change to clustercomponent name

PR #882, opened 2023-06-13
fix (proposed): set the default tfenv to linux for EC2 instances, local dev, etc.

kuannie1:
I'm not sure if this is the smartest change, but I think it could be a first step. For a lot of us with M1 macs, we run into incompatibilities where our tfenv architecuture is set to darwin_arm64 instead of the linux one. Most of our EC2 Instances run in amd though--we don't want to change the default to the newest Macbook default. 

For example, I'd get this error a lot when I run terraform locally:
```
╷
│ Error: Incompatible provider version
│ 
│ Provider registry.terraform.io/hashicorp/template v2.2.0 does not have a
│ package available for your current platform, darwin_arm64.
│ 
│ Provider releases are separate from Terraform CLI releases, so not all
│ providers are available for all platforms. Other versions of this provider
│ may have different platforms supported.
╵
```
I couldn't get rid of it until I modified the conditions of the tf_env install--I didn't care about getting the M1 Macbook version so I looked into ways to convince tf_env I'm using linux_amd64 architecture.

This change exports and sets a TFENV architecture to Linux by default--I have not tested whether we can override it back to arm64 or not.

### DIFF
--- a/testdata/auth0_provider_yaml/scripts/common.mk
+++ b/testdata/auth0_provider_yaml/scripts/common.mk
@@ -12,6 +12,7 @@ TFENV_DIR ?= $(REPO_ROOT)/.fogg/tfenv
 export PATH :=$(TFENV_DIR)/libexec:$(TFENV_DIR)/versions/$(TERRAFORM_VERSION)/:$(REPO_ROOT)/.fogg/bin:$(PATH)
 export TF_PLUGIN_CACHE_DIR=$(REPO_ROOT)/.terraform.d/plugin-cache
 export TF_IN_AUTOMATION=1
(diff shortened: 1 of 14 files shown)


jakeyheath:
I think the real problem here is that some of our old modules utilized `registry.terraform.io/hashicorp/template`. I actually fixed it so that the module is updated, but some workspaces still reference the old one. Their state's need to be refreshed with the latest module before that disappears entirely. Here's the PR that I used to fix it:

https://github.com/chanzuckerberg/shared-infra/pull/7716

jakeyheath:
I'm confused about the comment about EC2 instances. Why would that matter in this case?

PR #1101, opened 2025-03-21
chore: CCIE-4284 replace tibdex/github-app-token

wontonst:


wontonst:
gave it 

> replace all uses of tibdex/github-app-token with actions/create-github-app-token@v1. the inputs to the action have also changed. use private-key instead of private_key and use app-id instead of app_id

and it worked.

if another action was using app_id it would've incorrectly replaced it though. I wonder if the AI would be smart enough to undo that change.

